PR #95, opened 2015-01-02
Override storefront_product_search

SiR-DanieL:
Let users override the function storefront_product_search.
When you use some plugins (eg. [WooCommerce Product Search](http://www.woothemes.com/products/woocommerce-product-search/)) you will probably want to use that AJAX form in the header.

Now you have to remove the action and add another one with a different function.


jameskoster:
grazie!

SiR-DanieL:
Di nulla ;)

PR #101, opened 2015-01-13
Fix for Footer showing wrong year for copyright...

niravmehta:
get_the_date() pulls date from the current global post object. The date of the current post object may not be the actual current date. So using date() instead of get_the_date() to show correct current year.


jameskoster:
Thanks!

PR #103, opened 2015-01-16
Added function check for storefront_header_cart

marcfalk:
It should be possible to overwrite this function from a child-theme. It might be the case that someone needs more than just the link and the widget (or a different widget), or like me, need the classes to be different.


jameskoster:
Agree, thanks!

PR #125, opened 2015-02-05
Composite Products tweaks

franticpsyx:
Found a couple missing styles while testing the UI, nothing major but worth pulling in.


jameskoster:
gracias!

PR #127, opened 2015-02-09
Reorders the sidebar registrations such that the primary sidebar is registered first. Fixes #126.

### LINKED ISSUE #126: Default widgets showing in header widget area
Hi, I've set up Storefront twice in the past week and both time, the default widgets were added to the header widget area, which made the homepage template look rather odd. Can we make that this doesn't happen?

It was here (but I've already deleted them):

mattyza:
@jameskoster Merge as desired.

Also, we should rename `sidebar-1` at some stage, as this is going to cause issues with many sidebars being generated, especially with a tool such as WooSidebars.

### DIFF
--- a/inc/functions/setup.php
+++ b/inc/functions/setup.php
@@ -95,8 +95,8 @@ endif; // storefront_setup
  */
 function storefront_widgets_init() {
 	register_sidebar( array(
-		'name'          => __( 'Header', 'storefront' ),
-		'id'            => 'header-1',
+		'name'          => __( 'Sidebar', 'storefront' ),
+		'id'            => 'sidebar-1',
 		'description'   => '',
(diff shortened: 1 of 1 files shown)


jameskoster:
I changed this so that the header region appeared at the top of the list on the Appearance > Widgets screen. I feel that makes more sense.

At some stage? Why not now? xD

jameskoster:
Ah OK, having seen Jobs issue I understand :p

mattyza:
@jameskoster The header region being at the top means that, on new installs, the default widgets are added to the header widget area. ;)

Feel free to change the slug of the sidebar now. I'm fine with that. Just need to bear in mind that customers would lose their sidebar widgets.

I'd say leave it for now and see if anyone brings up the slug? Should be okay to leave as is, in the short term.

PR #129, opened 2015-02-10
Adds the correct page content hook in the homepage template inline docs.

danieldudzic:
( 'storefront_homepage_content' )


jameskoster:
gracias

PR #140, opened 2015-02-26
cute type conversion removal?

aj-adl:
Borderline pedantic but I just noticed it while browsing the code and thought it was an easy fix.


jameskoster:
ty :p

PR #165, opened 2015-04-15
Download languages files from Transifex and create .mo files

claudiosmweb:
Hey @jameskoster 

Now you have the new Gruntasks:

- `tx_update`: Will create the .pot file and update the resourse in Transifex.
- `mo`: Will pull the .po files from Transifex and create .mo files

I also changed some tasks:

- `dev`: Runs the `default` and `tx_update` tasks.
- `deploy`: Rusn the `mo` and `copy` tasks.

For now I did not have any commits to the translations files.
However I believe that we could use the .gitignore not to push translations (.mo and .po) and this way you will have them only at WordPress.org, when running the `copy` task.
But that part is up to you, I'm just suggesting :p

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,0 +1,8 @@
+[main]
+host = https://www.transifex.com
+
+[storefront-1.storefront]
+file_filter = languages/<lang>.po
+source_file = languages/storefront.pot
+source_lang = en
+type = PO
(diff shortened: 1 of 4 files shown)


jameskoster:
@claudiosmweb you are beautiful!

PR #173, opened 2015-05-01
Provide spacing for multiple my-account buttons

WPprodigy:
When there are more than one actions available on the account page for an order, the buttons are run together.

Before: http://cld.wthms.co/wwHD

After: http://cld.wthms.co/1hM9n


claudiosmweb:
+1

jameskoster:
Thanks @WPprodigy !

PR #175, opened 2015-05-05
Increase readability for debug info

WPprodigy:
As is, the debug logs are pretty hard to read, and require cursor highlighting. This is a simple patch to darken the text:)

Before: http://cld.wthms.co/1eXrC
After: http://cld.wthms.co/1diUi

### DIFF
--- a/inc/woocommerce/css/woocommerce.scss
+++ b/inc/woocommerce/css/woocommerce.scss
@@ -1209,6 +1209,10 @@ p.no-comments {
 			opacity: 0.7;
 		}
 	}
+
+	pre.debug_info {
+		color: #484c51;
+	}
 }
 
 .wc-forward {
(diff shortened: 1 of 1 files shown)


jameskoster:
Hey @WPprodigy good catch. Can I make a suggestion? I'd say make the text the same color as regular text in the messages, but make the background color a translucent black, something like `rgba(#000,.1);`

Update your branch and I'll merge.

Thanks again man! :-)

WPprodigy:
Yea, I agree with that. I made the decision to change the text based off of the FedEx Debug, as it oddly has some inline styles for background color: http://cld.wthms.co/siBV

But it seems FedEx Shipping is the only one that does this, with UPS and USPS both taking the default StoreFront styles: http://cld.wthms.co/siBV

So yep, I'll submit the patch here for a translucent black bg color, and then probably a patch to get rid of those inline styles for FedEx.

WPprodigy:
Removed the class declaration (.debug_info) as it was only present in FedEx debug logs. So now all <pre> text in the messages will be readable. 

After: http://cld.wthms.co/1dWYm

jameskoster:
Winning. Thanks @WPprodigy !

PR #178, opened 2015-05-07
Homepage featured products add order & orderby parameters #177

### LINKED ISSUE #177: Add functionality to reorder the output of homepage featured products
Adding the ability to reorder the homepage featured products via the filter and the order and orderby parameters would be cool. This would allow as an example the ability to easily randomize the featured products and the parameters are ones that area available in the WC shortcode.

stuartduff:
closes #177

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,8 +1,5 @@
 # Storefront Changelog
 
-### *2015.05.07* - 1.4.6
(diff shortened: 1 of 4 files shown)


jameskoster:
was unable to merge this so I made the change manually. Cheers @stuartduff !

PR #179, opened 2015-05-08
Check to see if "Layout_Picker_Storefront_Control" class exists

niravmehta:
This will allow child themes to override Layout options. (we created a child theme that offers a 1 column layout, so had to override this class)


jameskoster:
thanks!

PR #200, opened 2015-07-16
Fix escaping on 404

fklein-lu:
- Static strings (as in `echo 'hey';`) or numbers do not need to be
escaped, as there is no possibility to change them without changing the
code.
- When escaping numbers passed to query functions, `absint()` should be
used instead of `intval()`, to make sure to obtain positive integers.
- Strings that are output in HTML should be escaped with `esc_html()`
instead of `esc_attr()`.
- `esc_attr( __() )` can be written as `esc_attr__()`, same for
`esc_html( __() )`.
- Translations should always be escaped.

### DIFF
--- a/404.php
+++ b/404.php
@@ -16,10 +16,10 @@ get_header(); ?>
 				<div class="page-content">
 
 					<header class="page-header">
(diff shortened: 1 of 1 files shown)


jameskoster:
gracias

PR #210, opened 2015-07-24
Update template-tags.php

ardalann:
HTML Error Fix: Element ```div``` not allowed as child of element ```ul```.


jameskoster:
Thanks! :-)

PR #211, opened 2015-07-31
Updated styles for CP 3.2 and PB 4.11 compatibility

franticpsyx:
Since CP 3.2 and PB 4.11, I have decided to apply the positive-wrapper-padding / negative-image-margin trick in order to have selected/bundled item images in a fixed-width left column, and text filling up the remaining space.

I've made some slight tweaks to the Storefront css in order to set those paddings/margins back to zero and allow Storefront to apply its own grid styles. Also fixed a couple issues with hidden thumbnails.

In case you are wondering why I'm not using media queries + percent-width columns: I need the total available container width to style the columns properly, not the viewport width. (Some themes place product forms in a column next to the main image, and others allow stacking the product form under the main image.) So the safest, simplest approach is to have a small, fixed-width thumbnail and let text take up the remaining space.

(On a sidenote, CP relies on adding/removing classes via JS to change the style of its responsive product loop thumbnails based on a container width measurement on init/resize. This technique, however, is only used in CP and it is limited to the component option loops + summary sections, so far with great success.)

I've tested these minor changes to make sure they are backwards compatible. If you want to take a look, check out the `development` branches of both extensions.

EDIT: Both extensions went live a few hours ago.


jameskoster:
Eep, we're no longer tracking the compiled .css files so I can't merge this @franticpsyx. Would you mind updating your PR to only include the .scss files? I'll then merge.

franticpsyx:
Incoming!!!

franticpsyx:
@jameskoster Done!

jameskoster:
muchos gracias!

PR #215, opened 2015-08-17
Wrap the place order button on line breaks so buttons don't wrap on characters.

### LINKED ISSUE #214: Place Order Button Wraps Incorrectly
Slightly longer button titles provided by gateways can wrap in the wrong place. For example "Request Confirmation" from the WC Bookings plugin can cause the following to happen with storefront:

justinshreve:
Fixes #214.

Slightly longer button titles provided by gateways can wrap in the wrong place. For example "Request Confirmation" from the WC Bookings plugin was causing the following to happen:

This tiny PR tells the button to wrap on line breaks when possible to avoid that from happening:

### DIFF
--- a/inc/woocommerce/css/woocommerce.scss
+++ b/inc/woocommerce/css/woocommerce.scss
@@ -838,6 +838,7 @@ form.checkout {
 		.button {
 			font-size: 1.387em;
 			width: 100%;
+			white-space: pre-wrap;
 		}
 	}
 }


mikejolley:
@mattyza someone can review this while James is afk?

mattyza:
@mikejolley @justinshreve Merged. Will run grunt tasks now.

mattyza:
@mikejolley @justinshreve FYI, this will go into the next release of the theme. Will submit alongside the next release we submit to WP.org.

PR #219, opened 2015-08-19
Better enqueue RTL resources.

georgestephanis:
Instead of toggling the stylesheet that’s enqueued, just tell Core where both of them are, and let Core handle the swapping in a ‘late-binding’ fashion.

Relevant core code:
https://github.com/WordPress/WordPress/blob/809baf442b/wp-includes/class.wp-styles.php#L88-L104

Also, if Storefront ever later migrates to shipping minified versions of these css in addition the normal css sources, you can specify a .min suffix like this:

```php
wp_style_add_data( 'example', 'suffix', $min );
```

### DIFF
--- a/inc/functions/setup.php
+++ b/inc/functions/setup.php
@@ -143,11 +143,9 @@ function storefront_widgets_init() {
 function storefront_scripts() {
 	global $storefront_version;
 
-	if ( is_rtl() ) {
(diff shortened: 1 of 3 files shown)


jameskoster:
This is awesome, thanks @georgestephanis !

PR #220, opened 2015-08-25
Check for featured products on 404 page

WPprodigy:
Prevent this: http://cld.wthms.co/1aptw/1UlXRHz6

By turning it into this: http://cld.wthms.co/SIJz/1DstbzTT

Could also do a check to see if products are on sale and show that before reverting to recently added.

### DIFF
--- a/404.php
+++ b/404.php
@@ -36,23 +36,7 @@ get_header(); ?>
 
 							echo '<div class="col-1">';
 
-								if ( wc_get_featured_product_ids() ) {
-
-									echo '<h2>' . esc_html__( 'Featured Products', 'storefront' ) . '</h2>';
-
-									echo storefront_do_shortcode( 'featured_products', array(
-																'per_page' 	=> 2,
-																'columns'	=> 2,
-															) );
-								} else {
-
-									echo '<h2>' . esc_html__( 'Recently Added Products', 'storefront' ) . '</h2>';
-
-									echo storefront_do_shortcode( 'recent_products', array(
-																'per_page' 	=> 2,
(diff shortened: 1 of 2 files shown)


WPprodigy:
On second glance, could also make a storefront function that checks behind the scenes for featured products, then sale products. If one of them is found, then the function would output the correct shortcode. If no featured or sale products are found, then it would revert to recently added.

This could help clean up the code/logic needed in the template. Just a thought :)

jameskoster:
> On second glance, could also make a storefront function that checks behind the scenes for featured products, then sale products. If one of them is found, then the function would output the correct shortcode. If no featured or sale products are found, then it would revert to recently added.

Nice idea. You wanna work on this @WPprodigy or happy for me to take over?

WPprodigy:
I can give it a go:)

jameskoster:
Any idea when you plan to do this @WPprodigy ? Would like to get a release out this week.

WPprodigy:
How's that? Tried to make it so the function can be extended and used elsewhere.

jameskoster:
Looks good @WPprodigy. Couple of tiny last points;

* I reckon we can make the function name a little nicer, something like `storefront_promoted_products()`?.
* Maybe add a `$recent` arg, so there's an option to disable the recent products fall back?
* 'Products On Sale' can just be 'On Sale Now'
* 'Recently Added Products' can be 'New In Store'.

WPprodigy:
I agree on the function name. Was struggling to name it haha xD

jameskoster:
Perfection! Thanks once again :-)

PR #227, opened 2015-09-07
Fix - Pass the path inside site_url for customizer args link

shivapoudel:
@jameskoster Below are the fix and tweaks I have maintain:

1. Used `add_query_arg()` for customizer return and url args.
2. Use of `autofocus%5Bpanel%5D=nav_menus`  to configure menu panel focus.

Trailingslash was missing so only spinner was loading.
`site_url( '?storefront-customizer=true' )` instead of `site_url() . ?storefront-customizer=true`


jameskoster:
Thanks a bunch @shivapoudel ! :-)

PR #235, opened 2015-09-23
Add support for tabular bundles mini-extension

franticpsyx:
https://github.com/somewherewarm/woocommerce-product-bundles-tabular-layout - mimics the Grouped Products layout (table).

Current Storefront integration applies some rules that break the tabular layout, so I've retargeted those using `div` selectors (become `tr` when applying the tabular layout, no need to apply any custom styles there).


jameskoster:
Thanks :)

PR #241, opened 2015-09-30
removed comma to validate json and fix npm install error

jonharmon:
I believe that comma needs to come out to fix a json parse error i get when I do npm install. As soon as I removed it I could complete npm install.


jameskoster:
Thanks @jonharmon :-)

Issue #245, opened 2015-10-09
Styles for the new password strength meter


claudiosmweb:
Now we have a password strength meter and this need styles on storefront:
https://github.com/woothemes/woocommerce/pull/9319

In core we use:

```scss
	/**
	 *  Password strength meter
	 */
	.woocommerce-password-strength {
		text-align: center;
		font-weight: 600;
		padding: 3px 0px 3px 0px;

		&.strong {
			background-color: #c1e1b9;
			border-color: #83c373;
		}

		&.short {
			background-color: #f1adad;
			border-color: #e35b5b;
		}

		&.bad {
			background-color: #fbc5a9;
			border-color: #f78b53;
		}

		&.good {
			background-color: #ffe399;
			border-color: #ffc733;
		}
	}
```

claudiosmweb:
Default themes:
![screenshot from 2015-10-09 14 16 23](https://cloud.githubusercontent.com/assets/1264099/10400658/66d164a0-6e90-11e5-8941-3918793c0f01.png)

Storefront:
![screenshot from 2015-10-09 14 16 39](https://cloud.githubusercontent.com/assets/1264099/10400657/66d166ee-6e90-11e5-9bc5-91f0aacc152b.png)

nishitlangaliya:
@claudiosmweb, ok will implement this style for storefront themes. Asap and will commit code.

nishitlangaliya:
@claudiosmweb , fixes pushd and done PR for review.

PR #248, opened 2015-10-14
color code applied to match the storefront theme ref #245

nishitlangaliya:
@jameskoster, Hello Color code applied to .woocommerce-password-strength class for match the style of storefront theme. please review it when you get a chance and let me know your thoughts for the same.

Thanks. :)


jameskoster:
This should have been applied to your previous PR. nvm, I will fix :-)

PR #254, opened 2015-10-24
Add missing hooked function

WPprodigy:
Should these have the same priority?

https://github.com/woothemes/storefront/blob/master/inc/structure/hooks.php#L76


jameskoster:
@WPprodigy No, comments should have priority 20. Wanna update your PR? :-P

WPprodigy:
Oh? How come? They both show priority 10 here: https://github.com/woothemes/storefront/blob/master/inc/structure/hooks.php#L75-L76

I would think comments should be moved up to 20 though?

jameskoster:
> How come?

Because if they both have the same priority nothing can be added between them.

Think this was just an oversight when this code was originally written.

WPprodigy:
Oops, sorry about that earlier comment. I misread your post and thought you were saying that just the commented line should say 20 ;p

Thought I deleted it in time though, so now there is just an awkward gap in conversation

jameskoster:
Too much Mt. Dew.

Thanks man xD

PR #260, opened 2015-10-29
Improve grunt workflow

shivapoudel:
@jameskoster Following improvements:

* Exclude markdown files for deploy.
* Check linting errors before minification of scripts.
* No need of `rtlcss` in `grunt dev` as its included by default.
* In `grunt deploy` there exist no `mo` task so let's add `dev` tasks ;)

Simple!


jameskoster:
Thanks @shivapoudel ! :-)

Issue #269, opened 2015-11-06
stock color


jameskoster:
in stock = green
out of stock = red

Inherit colors from messages.

nishitlangaliya:
@jameskoster, should I try and do the pull request for the same ?

jameskoster:
@nishitlangaliya you can if you want :)

nishitlangaliya:
@jameskoster, Sure Thanks will update you for the same... :)

nishitlangaliya:
@jameskoster , I have applied the color code to in stock and out stock message and done the pull request https://github.com/woothemes/storefront/pull/270 for the same. please review it when you get a chance 

Thanks :)

nishitlangaliya:
@jameskoster , I am still getting message during pull request 'can’t automatically merge' though I had taken fresh clone get pull from master and created branch from it. do I need to delete all my old branch  ?

jameskoster:
@nishitlangaliya yes you will probably have to start again, you can't have had the latest scss files.

nishitlangaliya:
@jameskoster Thanks for reply . I will do the same as you sugeested and will do the PR.

nishitlangaliya:
@jameskoster  I had done the pull request https://github.com/woothemes/storefront/pull/272 without any conflict please check for the same.

Thanks :)

tiagonoronha:
@nishitlangaliya this looks good, thanks! :) Just a suggestion, to make the SASS easier to follow, perhaps change it to:

```
&.in-stock {
	color: $success;

	&:before {
		content: "\f118";
	}
}

&.out-of-stock {
	color: $error;

	&:before {
		content: "\f119";
	}
}
```

After that I'll be happy to merge this. :)

nishitlangaliya:
@tiagonoronha  Thanks for your suggestions as per your suggestion I have done the changes and done PR - https://github.com/woothemes/storefront/pull/274 for the same. can you please review it when you geta a chance :)

tiagonoronha:
Looks good, merged!

Thanks, @nishitlangaliya.

nishitlangaliya:
@tiagonoronha Thanks  :)

PR #270, opened 2015-11-06
color code applied to in stock and out stock message ref #269

nishitlangaliya:


jameskoster:
There are conflicts here so I cannot merge. Remember to `git pull` before working.

Also, you shouldn't need to apply the color the the `:before` and `:after` elements.

nishitlangaliya:
@jameskoster, Ok sure will do and let you know for the same.

Thanks

PR #271, opened 2015-11-06
fixes:conflict stock color ref #269

nishitlangaliya:


nishitlangaliya:
@jameskoster  would you please let me know is there any still conflict . I have already taken new clone then get the pull from master then created branch from it.

PR #276, opened 2015-11-10
Composite Products integration tweaks

franticpsyx:
- Styles the 'Clear selections' button in a way similar to `a.reset_variations`:

ref: http://cl.ly/image/3D0X3L0U2l0k

- Fixes an issue with padding + div elements wrapped under `component_selections`. The existing rule does not remove the left padding from all child divs and in a couple instances incorrectly removes the bottom padding.


jameskoster:
Thanks @franticpsyx ! :-)

PR #284, opened 2015-11-26
Adds logic to adjust to full-width layout if the main sidebar doesn't contain any widgets.

mattyza:
When it comes to the question of sidebars and widgets in themes, there seem to be 2 general approaches:

1. Display some text in the widget area, encouraging the website administrator to add widgets (*cringe*).
2. Leave a blank space where the widgets should be (no direct education for the website administrator about what elements are missing).

This pull request introduces a third option, where by we adjust the layout to be full width if no widgets are present in the sidebar.

As widgets are added by default to the main sidebar, we can make a few assumptions:
1. If the website administrator doesn't want widgets in the sidebar, they'll remove these widgets.
2. If the widgets are removed, and the sidebar remains empty, the website administrator doesn't want a sidebar.
3. Not many website administrators enjoy a large gap of white space next to their content. :)

While a small change (3 lines of code), I feel this change would make a significant impact in assisting store owners to get up and running as quickly as possible, without the need to a) write code to get rid of the large blank space (which they'll most likely do using CSS in a hack-y manner, which disrupts our grid), or to purchase our WooCommerce Customizer extension for Storefront and begrudgingly spend money with us for one small feature which the theme should provide out of the box. While we want folks to buy extensions from us, this should be a purchase made willingly, not begrudgingly.

Thoughts? :)

@jameskoster @tiagonoronha @jeffikus @stuartduff @ianstewart

### DIFF
--- a/inc/functions/extras.php
+++ b/inc/functions/extras.php
@@ -54,6 +54,11 @@ function storefront_body_classes( $classes ) {
 		$classes[] = 'storefront-cute';
 	}
 
+	// If our main sidebar doesn't contain widgets, adjust the layout to be full-width.
+	if ( ! is_active_sidebar( 'sidebar-1' ) ) {
+		$classes[] = 'storefront-full-width-content';
+	}
+
 	return $classes;
 }
 


jameskoster:
As mentioned in Slack, I'm not really a fan of this for a couple of reasons;

* Yeah, the empty space in sidebar might not be considered visually appealing (I wouldn't explicitly say that is _unappealing_ though) but this is the standard behaviour of most themes when no widgets have been set and I'd expect users to be familiar with it. It's also pretty rare that anyone would encounter a site without any widgets unless they removed them themselves. I don't think that anyone _expects_ removing all widgets to remove the entire sidebar and adjust the layout.
* Most importantly this can lead to potentially confusing/bad habits. Users will learn that if they want a full width layout they need to avoid using widgets. So if they want full width product pages they'd remove all widgets. Then, say they want layered nav on their archives, they add the widget, and their single product pages are no longer full width. It puts them in a catch-22 position that we should avoid. I can see this causing support tickets for sure.
* This is already a feature of our various extensions. If folks want full width product pages then they should use the WooCommerce Customiser extension. That will allow them to have full width products and standard width everything else.

If this is really a big issue I'd rather just release some small plugins; 'Storefront Full Width Product Pages', 'Storefront Full Width Product Archives', etc. That gives us a very simple canned response to anyone who needs this particular feature, and keeps it totally modular.

I know folks are averse to touching code, but this is really _really_ a tiny customisation, we could document it as well.

mattyza:
Thanks @jameskoster. :)

> ...but this is the standard behaviour of most themes when no widgets have been set and I'd expect users to be familiar with it.

I don't believe this is a standard, or that the standard has been set by anyone. Additionally, if it's a standard, that doesn't make it necessarily correct or incorrect. :)

> It's also pretty rare that anyone would encounter a site without any widgets unless they removed them themselves.

If widgets are present, the layout would remain as is. If I remove widgets from the sidebar, I expect my next step would be to have to take care of that large empty space. I'm certain there isn't a single soul who would like an empty space next to their content. In my view, it's about assisting store owners with the design elements, and ensuring they have a perfect design at all times, regardless of the buttons they click.

Having a huge empty space to the side, for a non-technical person, may cost a significant amount of time and money to resolve. Meanwhile, it's 3 lines of code for us. :)

> I don't think that anyone expects removing all widgets to remove the entire sidebar and adjust the layout.

I do. :) The design looks incorrect, otherwise.

> That might make sense to us, but one thing I've learned is that most people expect all themes to behave in the same way in regards to things like this.

> Most importantly this can lead to potentially confusing/bad habits. Users will learn that if they want a full width layout they need to avoid using widgets. So if they want full width product pages they'd remove all widgets. Then, say they want layered nav on their archives, they add the widget, and their single product pages are no longer full width. It puts them in a catch-22 position that we should avoid. I can see this causing support tickets for sure.

We already provide an array of ways to control widget visibility per-page or view (archives, etc). I'd say we end up with 3 camps of users; "I want widgets to be the same everywhere", "I want no widgets", and "I want absolute control over the widgets on every screen of my website". Those in the third camp will have already found their preferred solution, so we can assume they are fully taken care of (the user in your example would have already found a way to display the layered navigation only on their archive screens).

Users in the third camp aren't the concern of this pull request. This aims to focus on the first and second camps of users ("I want widgets" and "I don't want widgets").

For anyone who wants widgets, this is a non-breaking change. For those who don't want widgets, they can rejoice in that they don't have to spend time and money to write 3 lines of code they don't know how to write. It gets them selling quicker.

> This is already a feature of our various extensions. If folks want full width product pages then they should use the WooCommerce Customiser extension. That will allow them to have full width products and standard width everything else.

We don't need to nickel and dime our users. As mentioned in the PR description, these users should be purchasing the WooCommerce Customiser extension willingly and because of all the features it offers, not to take care of what I'd say is a design and layout error, in my opinion. Having to begrudgingly purchase a product isn't a great footing upon which to start a relationship with us.

> If this is really a big issue I'd rather just release some small plugins; 'Storefront Full Width Product Pages', 'Storefront Full Width Product Archives', etc. That gives us a very simple canned response to anyone who needs this particular feature, and keeps it totally modular.

A plugin for 3 lines of code hardly seems worthwhile, especially when it's not a logical step for the store owner to take. "To fix a theme-related visual issue, I need a plugin". That doesn't make sense, to me. :)

> I feel it at least needs more research before we just go ahead and implement it, anyway.

As a non-breaking change, I reckon it's pretty safe to implement and assess.

jameskoster:
Just because it's non-breaking that doesn't mean it wont cause issues. See my header image example. Our implementation there goes against the standard and I see lots of support queries as a result. In 12+ months I've never seen a single person bring up the idea of making the layout full width when no widgets are added to the sidebar.

Overall I'm just not comfortable with leading people to associate the absence of widgets with having an impact on the layout. That can only lead to more questions/support. 

> I've removed all widgets to achieve a full width layout on my product pages. How do I now have widgets on my product archives, pages, blog posts, blog archives, etc? 

Sure, we can recommend WooSidebars, but then you're recommending a widget management plugin (and a lot of relative overhead) to address what is fundamentally a layout issue and what should be 5-6 line customisation, or a tiny plugin. Surely it would be better in fact to intercept this a step earlier, when folks ask how to make their product pages full width, and recommend our purpose built extension which does this and much more (WooCommerce Customiser)?

Yes this would be non-breaking and solve an edge case, but could have a negative butterfly effect overall so needs to be thoroughly considered.

Also - one thing this doesn't account for is whether you've hooked a non-widget component in to the sidebar. If you have, but have no widgets, then the thing you expect to appear in the sidebar adjacent to your content now appears beneath it.

Interested to hear what everyone else things.

mattyza:
> The widget area should still be there, irrespective of whether there are actually any widgets in it or not.

While I'm pretty sure you're referring to the open space in this context, it brings up an interesting point.

https://github.com/woothemes/storefront/blob/master/sidebar.php#L9

In `sidebar.php`, the `#secondary` DIV (which we specifically designate as the container in which we display things to fill the gap next to the content), is entirely hidden if no widgets are present in the sidebar.

By the above statement, it would make sense to have the empty DIV tag, with an action hook inside it, rather than having the hooked code have to recreate `#secondary` each time. That could lead to conflicts where multiple code snippets attempt to recreate `#secondary`.

Point being, the logic we already have in Storefront hides the entire `#secondary` DIV, which indicates that we don't want the DIV there if there are no widgets in the sidebar (ie: we're not giving anyone an opportunity to hook in at this point in the first place, so "what if they hook in" is moot). :)

@jeffikus @stuartduff @tiagonoronha would love y'all's thoughts on this. I see both points here, and would love to unpack this discussion further, as I feel it's fruitful. :)

mattyza:
For what it's worth, the Twenty Sixteen theme adapts it's content width based on whether or not widgets are present in the sidebar. :)

@jameskoster

jeffikus:
I'm of the opinion that designs should be adaptive based on the content to be honest. This came up when I did a bug fix recently on Jotter for the Contact Form page template - see https://github.com/woothemes/jotter/issues/12 - and we discussed this at length in Slack. TL;DR the design was intended to not have a sidebar so it should be full width instead of having a blank space.

> ...but this is the standard behaviour of most themes when no widgets have been set and I'd expect users to be familiar with it.
I would prefer that a design became full width instead of having a blank sidebar.

But again the issue here is what is the intended "design" - and this is subjective. There is no set rule here. 

For me, I would think that a design should adapt within the bounds of the wrapper. Whether that wrapper is full width or boxed doesnt matter. 

In this case, my opinion is that if there is no sidebar present, it should then be full width within the bounds of the #content div. That's just my opinion though.

cc @mattyza @jameskoster @tiagonoronha

jameskoster:
Thought about this a little more over the weekend and I'm coming around the idea a little bit. As long as we don't actually recommend this to people who want a full width layout for one specific section of their site.

Will discuss with @tiagonoronha in our catchup today and make a final decision.

jeffikus:
:dancer: nice to see this merged!

PR #360, opened 2016-03-27
Add automatic child theme handling

mgibbs189:
This class ensures that a child theme is installed and active. It helps ensure that people are using a child theme instead of the parent.

Reference: https://github.com/FacetWP/use-child-theme

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -17,6 +17,7 @@ $storefront_version = $theme['Version'];
 require 'inc/class-storefront.php';
 require 'inc/jetpack/class-storefront-jetpack.php';
 require 'inc/customizer/class-storefront-customizer.php';
+require 'inc/use-child-theme.php';
 
 require 'inc/storefront-functions.php';
 require 'inc/storefront-template-hooks.php';
--- a/inc/use-child-theme.php
+++ b/inc/use-child-theme.php
@@ -2,10 +2,8 @@
 /*
  * Use Child Theme
  * A drop-in to make it easy to use WordPress child themes
- * @version 0.3
+ * @version 0.4
  */
-
-defined( 'ABSPATH' ) or exit;
 
 if ( ! class_exists( 'Use_Child_Theme' ) ) {
 
@@ -52,9 +50,7 @@ if ( ! class_exists( 'Use_Child_Theme' ) ) {
         (function($) {
             $(function() {
                 $(document).on('click', '.uct-activate', function() {
-                    $.post(ajaxurl, {
-                        action: 'uct_activate'
-                    }, function(response) {
+                    $.post(ajaxurl, { action: 'uct_activate' }, function(response) {
                         $('.uct-notice p').html(response);
                     });
                 });
@@ -74,7 +70,7 @@ if ( ! class_exists( 'Use_Child_Theme' ) ) {
 
 
         function dismiss_notice() {
-            set_transient( 'uct_dismiss_notice', 'yes', 86400 );
+            set_transient( 'uct_dismiss_notice', 'yes', apply_filters( 'uct_dismiss_timeout', 86400 ) );
             exit;
         }
 
@@ -135,9 +131,6 @@ if ( ! class_exists( 'Use_Child_Theme' ) ) {
         }
 
 
-        /**
-         * Generate the style.css contents
-         */
         function style_css() {
             ob_start();
 ?>
@@ -152,9 +145,6 @@ Version:        1.0
         }
 
 
-        /**
-         * Generate the functions.php contents
-         */
         function functions_php() {
             ob_start();
 ?>
--- a/inc/use-child-theme.php
+++ b/inc/use-child-theme.php
@@ -1,0 +1,173 @@
+<?php
+/*
+ * Use Child Theme
+ * A drop-in to make it easy to use WordPress child themes
+ * @version 0.3
+ */
+
+defined( 'ABSPATH' ) or exit;
+
+if ( ! class_exists( 'Use_Child_Theme' ) ) {
+
+    class Use_Child_Theme
+    {
+
+        public $theme;
+        public $child_slug;
+
+
+        function __construct() {
+            add_action( 'admin_init', array( $this, 'admin_init' ) );
+        }
+
+
+        function admin_init() {
+
+            // Exit if unauthorized
+            if ( ! current_user_can( 'switch_themes' ) ) {
+                return;
+            }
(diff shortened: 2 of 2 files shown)


BurlesonBrad:
...eh... not that I'm anyone of any 'decision making' level, but @mgibbs189 is this something that SF could do on a 'universal' level w/out causing folks **craniums to quiver?!?!?** :scream:  

In other words, if SF is growing as a 'base theme', and Woo encourages child themes for both financial and stylistic reasons, would people be totally CORNFUSED if they saw something like this:
sf-child-theme-maker

again, I'm just a dude who found a keyboard, and I've got no decision making powers. But, I'd be willing to bet that if this was actually implemented into Storefront, every WooNinja would loath this feature so much that they'd hunt you down and troll your interwebs until 
- [ ] they infected your computer with Windows XP, 
- [ ] infiltrated your cell phone with Nokia software, 
- [ ] bombarded your WordPress with base64 emoji poo spam, 
- [ ] and, then ban you off from all WordCamps until version WP 26.1 rolls out

_...but then again, I could be totally wrong as well!_ :stuck_out_tongue_winking_eye:

seb86:
@mgibbs189 I too think this is wrong. Storefront is just as good as a theme on it's own.

If you are trying to help the user create a child theme automatically for them so that they can then apply customizations to it. Then you need to address this properly and not have an admin notice that is telling the user that it is now required to use a child theme once Storefront has been activated.

A child theme, in my personal opinion, has always been optional.

If any customization needs to be applied then by all means recommend to use a child theme but not force feed something that is not required upon activating the theme.

Please undo this commit or remove the admin notice and place the action link elsewhere. Thank you.

BurlesonBrad:
he doesn't need to "undo this commit". @mikejolley @jameskoster @claudiosmweb or someone else @ WooHQ who is an owner with "write access" will take care of it :+1:

mgibbs189:
@BurlesonBrad pitchforks away, this is simply a conversation starter ;)

@seb86 What exactly is the "proper" way of handling this then? The user sees a notice. If they click Activate, it'll automatically create a child theme and switch to it. If they dismiss it, it will be hidden for a set time (default 1 day). Also, this isn't a commit, it's a pull request...

IMO users should be strongly encouraged to use a child theme. If a child theme is already in use, then the notice will never appear.

It's in Storefront's best interest to ensure that the parent theme isn't modified directly.

seb86:
@mgibbs189 I did not know the user could dismiss the notice from the screenshot that @BurlesonBrad had put. If that is possible then thats fine but I still think that the wording needs tweaking. 

**Suggestions**

1. Don't use the error notice. Use the information notice instead.
2. Also, give the user the ability to remove the notice indefinitely should they choose. Otherwise the shop owner is removing it each day when they log in.

BurlesonBrad:
@seb86 look at line 73 https://github.com/FacetWP/use-child-theme/blob/master/use-child-theme.php#L73
Once the end-user dismisses the notification the first time, the notification will **continue** to **endlessly** appear until the end-user eventually clicks it, or opens a support ticket asking what it means and why it won't go away.

sorry @mgibbs189 I usually come with a sign around my neck that says "Must have sense of humor and/or thick skin around Brad".  Again, GREAT idea for a plugin though :+1:

jameskoster:
This is very interesting and I like the _general idea_. Thanks for opening up the conversation @mgibbs189 

A couple of points;

* As @seb86 pointed out the wording would need to change a bit. This doesn't explain _why_ the user should be using a child theme. There needs to be a bit more explanation imo.
* We don't generally recommend child themes for editing themes at Woo. The reason for that is twofold; 1. We offer a catalog of Storefront child themes already (free and commercial). Obviously you can't use a 'customisation' child theme and one of our child themes concurrently. 2. It ties your customisations to something that you might want to change in the future (IE you might want to purchase a Storefront child theme in the future at which point you'd lose your customisations). 
* Our preferred customisation method is [this plugin](https://github.com/woothemes/theme-customisations). It allows you to isolate your customisations and easily enable/disable them in an intuitive way.

So, what I'm thinking here would be to modify this notice to link to that plugin instead. I'm actually considering uploading that plugin to .org which would make the installation process via admin notice very smooth. Whether we give folks the option to install a child theme, or the plugin is up for debate.

jameskoster:
Having considered this some more, I'm not sure if it's worth including this in core. We have links to our Customisation plugin in functions.php and style.css, the two locations that folks are most likely to edit.

It could be confusing to be met with a notice about customisation immediately after activation.

I think I'll close this out for now, but will likely upload the customisation plugin to .org.

mgibbs189:
@jameskoster Agreed, the preferred way to customize is via a plugin (I <a href="https://facetwp.com/how-to-use-hooks/">wrote a post about it</a> too).

In my experience, most non-technical users won't bother installing a plugin or keeping file changes outside of the theme. This project's aim is to *help* alleviate that.

You're right that Storefront might not be the best use-case, given how its ecosystem revolves around child themes. But I think the general problem remains... there needs to be a way to keep users from modifying the themes directly.

jameskoster:
@mgibbs189 I totally agree this is a big problem, I'm just not sure that this is the correct time / place to educate. It seems unlikely to me that someone would dive straight into customisation immediately after activation. They're more likely to just dismiss the notice and then probably forget about it. Making the notice persistent will just lead to frustration imo.

Another thing that we haven't considered here is that it is actually perfectly OK to modify Storefront core (or any parent theme) and use it as a 'starter theme' if you know what you're doing. It'd be annoying to see this notice in that case.

Let's also consider the behaviour of folks who edit parent theme files without knowledge of the consequences... I don't know for certain but I'd be willing to bet that the vast majority of them perform their edits via `theme-editor.php` in the Dashboard. So how about we just include the notice there? That would make more sense and be much less invasive imo.

mgibbs189:
> Let's also consider the behaviour of folks who edit parent theme files without knowledge of the consequences... I don't know for certain but I'd be willing to bet that the vast majority of them perform their edits via theme-editor.php in the Dashboard. So how about we just include the notice there? That would make more sense and be much less invasive imo.

Yes, that's a really great idea.

So, assuming that the notice only appeared on `theme-editor.php` and the notice text was more descriptive, do you think this would be a viable solution?

jameskoster:
Yeah definitely. I think that for Storefront we'll recommend the installation of our theme-customisations plugin rather than a child theme though. I'll open a new issue for that just now.

PR #365, opened 2016-04-05
Fix path to the jetpack.css file.

alaczek:
It looks like the path to jetpack.css file changed in Storefront 2.0.


jameskoster:
Thanks @alaczek !

PR #417, opened 2016-06-13
Remove My Account and Cart link if pages do not exist

SiR-DanieL:
In the bottom bar, the My Account and Cart links show even if the pages do not exist.

This PR checks if there's an ID set in WooCommerce before to show them. It may not be used a lot, but it could. I have a use case already with a shop with a only one product and using One Page Checkout.


jameskoster:
Thanks!

PR #420, opened 2016-06-23
Added check for existence of secondary nav menu

titanic-fanatic:
If not using a secondary nav menu, you end up with an empty <nav> tag which
messes up the layout of the search bar which get pushed below the title/logo.
This also causes the header to be larger than desired and looks terrible.
This patch simply checks for the existence of a secondary menu before it outputs
the <nav> tags, thus not echoing an empty <nav> tag if you don't use a secondary
nav menu.

### DIFF
--- a/inc/storefront-template-functions.php
+++ b/inc/storefront-template-functions.php
@@ -213,18 +213,20 @@ if ( ! function_exists( 'storefront_secondary_navigation' ) ) {
 	 * @return void
 	 */
 	function storefront_secondary_navigation() {
-		?>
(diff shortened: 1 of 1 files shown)


jameskoster:
Thanks!

Interesting username, referring to the ship? Why so fanatical about it? xD

titanic-fanatic:
It's a long story. Suffice to say that I have a long standing interest and connection to the Titanic and I know almost every last detail about it ;-)

jameskoster:
Had a similar fascination as a kid xD

PR #423, opened 2016-07-06
Remove zero margins in bundles.scss

franticpsyx:
Just noticed this in the WC Bundles integration sass - is the zero-margin rule critical? 
Can't think of any reason to eliminate the default margins there.


jameskoster:
@franticpsyx Did you test?

PR #432, opened 2016-07-12
Check if functions exist

martinvd:
Check if `storefront_primary_navigation_wrapper` or `storefront_primary_navigation_wrapper_close` exists. They might be already defined in a child theme.

### DIFF
--- a/inc/storefront-template-functions.php
+++ b/inc/storefront-template-functions.php
@@ -763,18 +763,22 @@ if ( ! function_exists( 'storefront_post_thumbnail' ) ) {
 	}
 }
 
-/**
- * The primary navigation wrapper
- */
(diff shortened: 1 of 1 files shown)


jameskoster:
Hmm, we left these un-pluggable as it's probably not a great idea to tweak this markup. Are you trying to do something specifically?

martinvd:
I was overriding them completely like other functions in that file offer this possibility. So I thought those 2 functions might have been overlooked, hence my PR. I admit the functions are pretty simple and to work around them I could also just remove the hook from the `storefront_header` action. So feel free what you will do with this :)

jameskoster:
I suppose it doesn't hurt to make this pluggable. Devs should be overwriting functions responsibly.

PR #436, opened 2016-07-21
fixed clearfix mixin typo causing sass to fail

jwogrady:
I believe this is a typo.... unless I'm unaware of some strange SCSS operator. :-)


jameskoster:
Good catch!

Issue #437, opened 2016-07-23
Semantic HTML Issues 


opportus:
Hi @jameskoster 

Storefront is great and diserve a perfect [Nu HTML check](http://validator.w3.org/check?verbose=1&uri=http://sauceweb.net/index.php/2016/06/22/fantastic/).
I could PR a fix for multiple warnings and errors if you'd like...

But here are the **most** problematic issues:

### Structural Elements
There are **multiple** HTML5 structural elements such as `<nav>` and `<section>` missing headings. [**This hurts the pages semantics**](http://html5doctor.com/avoiding-common-html5-mistakes/). As resume very well HTML5 Doctor, the structural elements are **not** wrappers...

### Document Outline
It's messed up on too many pages... from `<h1>` it jumps to `<h3>` without passing by `<h2>`...
On single post, I have a "Navigation de l'article" (post navigation) as `<h2>` !! I believe you did that for screen reader ? But for SEO it's **very bad**.
Suppose you have a keyword which you want to wrap into a `<h2>`, which should be the **only** purpose of `<h*>` tags, the annoying "post navigation" will disolve your keyword impact, and this on **every** single post pages...
Same story about **all widget titles** which should be wrapped into a `<span>` or a `<div>`, at least from a SEO point of view, which is the most important for **any** Storefront site owner. If this is not possible, then I'd strongly recommend using `<h4>` instead of `<h3>` for all of them.

...can we do something about all that ?

jameskoster:
@opportus Yup, there's definitely a lot of stuff we can improve here. A PR would be very much appreciated.

opportus:
@jameskoster 

I'm preparing my first commit right now, but I need to know what do you think about `<div>` instead of `<h3>` for widget titles ?
Let me list the pros and cons about this switch:

**Pros**
- You'll can now say **Storefront is SEO optimized** because it helps search engines to focus on the real content. For me, as a Storefront user, making some money from my e-commerces powered by your theme, this benefit worth **infinitly**  the drawbacks listed below.

**Cons**
- I don't know much about screen readers but I guess it won't help them ? Though, from the few lines I've readed about them, they have other ways to navigate to the widgets than by headings.
- We have to rewrite the CSS concerning widget titles. And between the customizer and the new scss, it's not very clear to me yet how to proceed.

lkraav:
> I'm preparing my first commit right now, but I need to know what do you think about <div> instead of <h3> for widget titles ?

[...]

PR #439, opened 2016-07-26
Improved Semantic HTML

opportus:
Here is the PR that aim to fix various problems listed in issue #437.

Since there are many interconnected little things to modify, I'll try to proceed my changes as much progressively and coherently as I can... commit after commit.

@jameskoster, I know that you might not necessarely agree with some of my changes but it's more conveniant for me to discuss about them after I commit and then change it the way you want :)

### DIFF
--- a/404.php
+++ b/404.php
@@ -11,7 +11,7 @@ get_header(); ?>
 
 		<main id="main" class="site-main" role="main">
 
-			<section class="error-404 not-found">
+			<div class="error-404 not-found">
 
 				<div class="page-content">
 
@@ -22,44 +22,52 @@ get_header(); ?>
 					<p><?php esc_html_e( 'Nothing was found at this location. Try searching, or check out the links below.', 'storefront' ); ?></p>
 
 					<?php
-					if ( is_woocommerce_activated() ) {
-						the_widget( 'WC_Widget_Product_Search' );
-					} else {
-						get_search_form();
-					}
+          echo '<section aria-label="Search">';
+
+					  if ( is_woocommerce_activated() ) {
+						  the_widget( 'WC_Widget_Product_Search' );
+					  } else {
+						  get_search_form();
+					  }
+
+          echo '</section>';
 
 					if ( is_woocommerce_activated() ) {
 
 						echo '<div class="fourohfour-columns-2">';
 
-							echo '<div class="col-1">';
+							echo '<section class="col-1" aria-label="Promoted Products">';
 
 								storefront_promoted_products();
 
-							echo '</div>';
+							echo '</section>';
 
-							echo '<div class="col-2">';
+							echo '<nav class="col-2" aria-label="Product Categories">';
 
 							echo '<h2>' . esc_html__( 'Product Categories', 'storefront' ) . '</h2>';
 
 							the_widget( 'WC_Widget_Product_Categories', array(
 																			'count'		=> 1,
 							) );
-							echo '</div>';
+							echo '</nav>';
 
-						echo '</div>';
+            echo '</div>';
 
-						echo '<h2>' . esc_html__( 'Popular Products', 'storefront' ) . '</h2>';
+            echo '<section aria-label="Popular Products" >';
 
-						echo storefront_do_shortcode( 'best_selling_products', array(
-																			'per_page' 	=> 4,
-																			'columns'	=> 4,
-						) );
+              echo '<h2>' . esc_html__( 'Popular Products', 'storefront' ) . '</h2>';
+
+						  echo storefront_do_shortcode( 'best_selling_products', array(
+							  'per_page' 	=> 4,
+								'columns'	=> 4,
+              ) );
+
+            echo '</section>';
 					}
 					?>
 
 				</div><!-- .page-content -->
-			</section><!-- .error-404 -->
+			</div><!-- .error-404 -->
 
 		</main><!-- #main -->
 	</div><!-- #primary -->
--- a/footer.php
+++ b/footer.php
@@ -14,7 +14,7 @@
 
 	<?php do_action( 'storefront_before_footer' ); ?>
 
-	<footer id="colophon" class="site-footer">
+	<footer id="colophon" class="site-footer" role="contentinfo">
 		<div class="col-full">
 
 			<?php
--- a/footer.php
+++ b/footer.php
@@ -14,7 +14,7 @@
 
 	<?php do_action( 'storefront_before_footer' ); ?>
 
-	<footer id="colophon" class="site-footer" role="contentinfo">
+	<footer id="colophon" class="site-footer">
 		<div class="col-full">
 
 			<?php
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -125,41 +125,56 @@ if ( ! class_exists( 'Storefront' ) ) :
 		 *
 		 * @link http://codex.wordpress.org/Function_Reference/register_sidebar
 		 */
-		public function widgets_init() {
-			register_sidebar( array(
-				'name'          => __( 'Sidebar', 'storefront' ),
-				'id'            => 'sidebar-1',
-				'description'   => '',
-				'before_widget' => '<div id="%1$s" class="widget %2$s">',
-				'after_widget'  => '</div>',
-				'before_title'  => '<h5 class="widget-title">',
-				'after_title'   => '</h5>',
-			) );
-
-			register_sidebar( array(
+    public function widgets_init() {
+			$sidebar_args['header'] = array(
 				'name'          => __( 'Below Header', 'storefront' ),
 				'id'            => 'header-1',
-				'description'   => 'Widgets added to this region will appear beneath the header and above the main content.',
-				'before_widget' => '<div id="%1$s" class="widget %2$s">',
-				'after_widget'  => '</div>',
-				'before_title'  => '<h5 class="widget-title">',
-				'after_title'   => '</h5>',
-			) );
+				'description'   => 'Widgets added to this region will appear beneath the header and above the main content.'
+			);
+      
+      $sidebar_args['sidebar'] = array(
+				'name'          => __( 'Sidebar', 'storefront' ),
+				'id'            => 'sidebar-1',
+				'description'   => ''
+      );
 
-			$footer_widget_regions = apply_filters( 'storefront_footer_widget_regions', 4 );
+      $footer_widget_regions = apply_filters( 'storefront_footer_widget_regions', 4 );
+      
+      for ( $i = 1; $i <= intval( $footer_widget_regions ); $i++ ) {
+        $footer = sprintf( 'footer_%d', $i );
+        
+        $sidebar_args[ $footer ] = array(
+			    'name'        => sprintf( __( 'Footer %d', 'storefront' ), $i ),
+				  'id'          => sprintf( 'footer-%d', $i ),
+				  'description' => sprintf( __( 'Widgetized Footer Region %d.', 'storefront' ), $i )
+        );
+      }
 
-			for ( $i = 1; $i <= intval( $footer_widget_regions ); $i++ ) {
-				register_sidebar( array(
-					'name' 				=> sprintf( __( 'Footer %d', 'storefront' ), $i ),
-					'id' 				=> sprintf( 'footer-%d', $i ),
-					'description' 		=> sprintf( __( 'Widgetized Footer Region %d.', 'storefront' ), $i ),
-					'before_widget' 	=> '<div id="%1$s" class="widget %2$s">',
-					'after_widget' 		=> '</div>',
-					'before_title' 		=> '<h5>',
-					'after_title' 		=> '</h5>',
-					)
-				);
-			}
+      foreach ( $sidebar_args as $sidebar => $args ) {
+        $widget_tags = array(
+				  'before_widget' => '<div id="%1$s" class="widget %2$s">',
+				  'after_widget'  => '</div>',
+				  'before_title'  => '<span class="widget-title">',
+				  'after_title'   => '</span>'
+        );
+        
+        // Dynamically generated filter hooks. Allow changing widget wrapper and title tags. See the list below.
+        // 
+        // 'storefront_header_widget_tags'
+        // 'storefront_sidebar_widget_tags'
+        //
+        // 'storefront_footer_1_widget_tags'
+        // 'storefront_footer_2_widget_tags'
+        // 'storefront_footer_3_widget_tags'
+        // 'storefront_footer_4_widget_tags'
+        // ...and so on if you did add more footer widget regions via `storefront_footer_widget_regions` filter hook.
+        $filter_hook = sprintf( 'storefront_%s_widget_tags', $sidebar );
+        $widget_tags = apply_filters( $filter_hook, $widget_tags );
+
+        if ( is_array( $widget_tags ) ) {
+          register_sidebar( $args + $widget_tags );
+        }
+      }
 		}
 
 		/**
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -26,14 +26,15 @@ if ( ! class_exists( 'Storefront' ) ) :
 		 * @since 1.0
 		 */
 		public function __construct() {
-			add_action( 'after_setup_theme',     array( $this, 'setup' ) );
-			add_action( 'widgets_init',          array( $this, 'widgets_init' ) );
-			add_action( 'wp_enqueue_scripts',    array( $this, 'scripts' ),       10 );
-			add_action( 'wp_enqueue_scripts',    array( $this, 'child_scripts' ), 30 ); // After WooCommerce.
-			add_filter( 'body_class',            array( $this, 'body_classes' ) );
-			add_filter( 'wp_page_menu_args',     array( $this, 'page_menu_args' ) );
-			add_action( 'enqueue_embed_scripts', array( $this, 'print_embed_styles' ) );
-			add_action( 'wp_footer',             array( $this, 'get_structured_data' ) );
+			add_action( 'after_setup_theme',          array( $this, 'setup' ) );
+			add_action( 'widgets_init',               array( $this, 'widgets_init' ) );
+			add_action( 'wp_enqueue_scripts',         array( $this, 'scripts' ),       10 );
+			add_action( 'wp_enqueue_scripts',         array( $this, 'child_scripts' ), 30 ); // After WooCommerce.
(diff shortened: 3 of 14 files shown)


inc/storefront-template-functions.php
```diff
@@ -183,8 +183,8 @@ function storefront_site_branding() {
 	 */
 	function storefront_primary_navigation() {
 		?>
-		<nav id="site-navigation" class="main-navigation" role="navigation" aria-label="<?php esc_html_e( 'Primary Navigation', 'storefront' ); ?>">
-		<button class="menu-toggle" aria-controls="primary-navigation" aria-expanded="false"><span><?php echo esc_attr( apply_filters( 'storefront_menu_toggle_text', __( 'Menu', 'storefront' ) ) ); ?></span></button>
+		<div id="site-navigation" class="main-navigation" role="navigation" aria-label="<?php esc_html_e( 'Primary Navigation', 'storefront' ); ?>">
```
jameskoster: @opportus Seems strange to me to _not_ use the nav tag here. Won't that be harmful? If the lack of header is the problem wouldn't it be better to simply add a header with the `screen-reader-text` class?

inc/class-storefront.php
```diff
@@ -153,10 +153,10 @@ public function widgets_init() {
 					'name' 				=> sprintf( __( 'Footer %d', 'storefront' ), $i ),
 					'id' 				=> sprintf( 'footer-%d', $i ),
 					'description' 		=> sprintf( __( 'Widgetized Footer Region %d.', 'storefront' ), $i ),
-					'before_widget' 	=> '<aside id="%1$s" class="widget %2$s">',
-					'after_widget' 		=> '</aside>',
-					'before_title' 		=> '<h3>',
```
jameskoster: Posted about widget titles here; https://github.com/woothemes/storefront/issues/437#issuecomment-235200498

opportus: OK

inc/storefront-template-functions.php
opportus: > Won't that be harmful?

Was it harmfull to use `<div>` in this context before HTML5 ? :p
But we're in HTML5 now and something I'm not aware of maybe changed...
What I'm sure about is that if we're going to keep `<nav>` then we should better leave it like it was, don't add tricky hidden heading there.
Your turn for the final decision.

jameskoster: > Was it harmfull to use <div> in this context before HTML5

No because `nav` didn't exist / wasn't recognised then xD

I think we should leave it as `nav`, I also think we should include the header because;

1. It's good for screen readers considering we have multiple navs
2. Valid html feels good.

footer.php
```diff
@@ -14,7 +14,7 @@
 
 	<?php do_action( 'storefront_before_footer' ); ?>
 
-	<footer id="colophon" class="site-footer" role="contentinfo">
```
opportus: I'm not sure anymore about removing role attributes on the specific elements elements such as `<header>`, `<footer>`, `<nav>`, etc... is a good thing.
From https://www.w3.org/WAI/GL/wiki/Using_HTML5_nav_element: 

> However, in order to make the structure more accessible to user agents that support ARIA as well as ensuring that user agents that don't support HTML5 can also understand the structure, adding the ARIA role="navigation" is advised.

Which just contradict their own tool "Nu HTML Checker" throwing warnings saying:

> Element nav does not need a role attribute.

...

inc/class-storefront.php
```diff
@@ -125,42 +125,57 @@ public function setup() {
...
-					'description' 		=> sprintf( __( 'Widgetized Footer Region %d.', 'storefront' ), $i ),
-					'before_widget' 	=> '<aside id="%1$s" class="widget %2$s">',
-					'after_widget' 		=> '</aside>',
-					'before_title' 		=> '<h3>',
-					'after_title' 		=> '</h3>',
-					)
-				);
-			}
-		}
+    public function widgets_init() {
```
opportus: I hope you won't consider my approach to implement filters "over complicated for such a little method"...
But I did start from the principle that users don't need to access the name, id and description of sidebar args. So does it all make sense ?

Not sure to understand what's the problem with the tabulation... is that because we use diffrent tab setting on our editors ?

opportus: @jameskoster 

I hope you won't consider my approach to implement filters "over complicated for such a little method"...
But I did start from the principle that users don't need to access the name, id and description of sidebar args. So does it all make sense ?

Not sure to understand what's the problem with the tabulation... is that because we use different tab setting on our editors ?

inc/storefront-template-functions.php
opportus: > I also think we should include the header

You just decided to remove headings from sidebars for you know which good reason... :P

> Valid html feels good.

Agreed, but we could argue about what is "valid HTML"... If you did read my line note about the `role` attribute, even W3C doesn't know. 😆 
Also, `<nav>` element without heading actually do **not** even throw a warning on Nu HTML Checker.

> It's good for screen readers considering we have multiple navs

There is nothing else to do for helping them ? We can't help them differentiate `<nav>` with `aria*` attributes ? I don't know much about that...

footer.php
```diff
@@ -14,7 +14,7 @@
...
+	<footer id="colophon" class="site-footer">
```
opportus: I do put them back... since it won't hurt anything anyway.

inc/class-storefront.php
```diff
@@ -125,41 +125,56 @@ public function setup() {
...
-					'id' 				=> sprintf( 'footer-%d', $i ),
-					'description' 		=> sprintf( __( 'Widgetized Footer Region %d.', 'storefront' ), $i ),
-					'before_widget' 	=> '<div id="%1$s" class="widget %2$s">',
-					'after_widget' 		=> '</div>',
-					'before_title' 		=> '<h5>',
-					'after_title' 		=> '</h5>',
-					)
-				);
-			}
+				'description'   => 'Widgets added to this region will appear beneath the header and above the main content.'
```
jameskoster: Just noticed this isn't localised :o

inc/storefront-template-functions.php
jameskoster: > You just decided to remove headings from sidebars for you know which good reason... :P

Oh yeah. Derp :D OK, let's just leave the heading out then.

> We can't help them differentiate <nav> with aria* attributes 

Good idea, see this link; https://www.w3.org/WAI/GL/wiki/Using_ARIA_landmarks_to_identify_regions_of_a_page#Example_3

inc/class-storefront.php
jameskoster: Looks good, don't worry about the tabbing.

inc/storefront-template-functions.php
opportus: @jameskoster thanks for the link. We don't need to localize the `aria-label` ?

inc/class-storefront.php
```diff
@@ -277,6 +278,18 @@ public function body_classes( $classes ) {
...
+    /**
+     * Custom navigation markup template hooked into `navigation_markup_template` filter hook.
+     */
+    public function navigation_markup_template () {
+      $template  = '<nav id="post-navigation" class="navigation %1$s" role="navigation" aria-label="Post Navigation">';
+      $template .= '<span class="screen-reader-text">%2$s</span>';
+      $template .= '<div class="nav-links">%3$s</div>';
+      $template .= '</nav>';
+
+      return $template;
```
jameskoster: Could add a filter here

single.php
```diff
@@ -16,14 +16,6 @@
 
 			get_template_part( 'content', 'single' );
 
-			/**
-			 * Functions hooked in to storefront_single_post_after action
-			 *
-			 * @hooked storefront_post_nav         - 10
-			 * @hooked storefront_display_comments - 20
-			 */
-			do_action( 'storefront_single_post_after' );
```
jameskoster: We shouldn't remove an action.

opportus: It shouldn't be a problem, right ?

inc/class-storefront.php
```diff
@@ -155,7 +155,7 @@ public function widgets_init() {
         $widget_tags = array(
           'before_widget' => '<div id="%1$s" class="widget %2$s">',
           'after_widget'  => '</div>',
-          'before_title'  => '<span class="widget-title">',
+          'before_title'  => '<span class="h3 widget-title">',
```
jameskoster: I thought we were using `h5`s for widget titles? You can use `.gamma` class for h3 styling.

opportus: OK

opportus: Hmm...
After that I did point at Yoast website using `<span>` as widget titles, you said:

> For the headings, let's go with a div and make the args filterable so folks can easily change if they want.

Did I miss something ?

single.php
opportus: I got a little doubt and checked the article and comments structure of HTML5 Doctor - http://html5doctor.com/understanding-aside/
Which is the same as I've brought here, hopefully.

inc/storefront-template-functions.php
```diff
@@ -165,7 +165,7 @@ function storefront_site_branding() {
 			jetpack_the_site_logo();
 		} else { ?>
 			<div class="site-branding">
-				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
+				<div class="beta site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></div>
```
opportus: An interesting topic on MOZ: https://moz.com/community/q/h1-tag-as-logo-on-the-homepage

opportus:
If you don't already, you can test the new document outline of this branch and compare it with master... We can see already serious improvements.
https://gsnedders.html5.org/outliner/

opportus:
If you don't already, you can test the new document outline of this branch and compare it with master... We can see already serious improvements.
https://gsnedders.html5.org/outliner/

But one problem remains on single post pages...

What we want our main content to be on these pages ? The post itself right ? At this time, semantically and structurally, it's not really. You can see it here: https://gsnedders.html5.org/outliner/process.py?url=http://sauceweb.net/index.php/2016/06/21/hello-world/
The main or top level content is actually what is outside of the `<article>`.
The `<article>` element which is a **sectioning** tag will just push the post in a lower **subsection** of the main content...

We can see that in post categories we don't have this problem: https://gsnedders.html5.org/outliner/process.py?url=http://sauceweb.net/index.php/category/uncategorized/
Because category heading is not wrapped into anything else than `<main>` which put it at the highest level possible. Each post wrapped in `<article>` **sectioning** element makes perfect sense here.

Some good ressources:
http://html5doctor.com/outlines/
https://www.smashingmagazine.com/2011/08/html5-and-the-document-outlining-algorithm/

@jameskoster, what do you think we should do here, do you think same as me ?

inc/storefront-template-functions.php
jameskoster: Is there any schema we should add here?

jameskoster:
@opportus Hmm, so to fix this we need to restructure our single posts page so that the `header` is outside of the `article`? But that doesn't seem to be correct either as `article` is a sectioning element and needs a `header`.

Alternatively we could swap the `article` tag on single posts with a `div`? But that doesn't feel right semantically. We can add schema to help but... meh.

What are your thoughts? :p

opportus:
@jameskoster

> Hmm, so to fix this we need to restructure our single posts page so that the header is outside of the `<article>` ? But that doesn't seem to be correct either as `<article>` is a sectioning element and needs a header.

Yup, that wouldn't make sense.

We don't have to use `<article>` because it's a post. The logic behind `<article>` or `<section>` **sectioning** tags is to differentiate a particular section from the main content.

inc/class-storefront.php
jameskoster: Nope, nvm me, confused myself xD

jameskoster:
@opportus then let's just change the `article` in content-single.php to a `div`?

opportus:
OK. I'm sure it's unarguable much better.

inc/storefront-template-functions.php
opportus: About logo or branding, SEO plugin such as Yoast already take care of many things. Optionally, we could use shema.org for the item "site" on every pages. Check this method on my WooCommerce PR.

```php
public function generate_shop_data() {
 +    if ( ! is_shop() || ! is_front_page() ) {
 +      return;
 +    }
 +
 +    $json['@type']             = 'WebSite';
 +    $json['name']              = get_bloginfo( 'name' );
 +    $json['url']               = get_bloginfo( 'url' );
 +    $json['potentialAction'] = array(
 +      '@type'                => 'SearchAction',
 +      'target'               => get_bloginfo( 'url' ) . '/?s={search_term_string}&post_type=product',
 +      'query-input'          => 'required name=search_term_string'
 +    );
 +
 +    $this->set_data( apply_filters( 'woocommerce_structured_data_shop', $json ) );
 +  }
 +}

Comment (opportus) [inline inc/storefront-template-functions.php]: And i just notice that Yoast plugin also generate website structured data on front page. So not sure we need it ?

File: single.php (hunk 8 above)
Comment (jameskoster) [inline single.php]: I think we should leave the actions where they were, but then add _new_ actions inside the `article`. `storefront_single_post_top`, `storefront_single_post_bottom`. Then hook those functions into the new actions instead.

File: inc/storefront-template-functions.php [hunk 11]
```diff
@@ -159,14 +159,20 @@ function storefront_header_widget_region() {
 	 * @return void
 	 */
 	function storefront_site_branding() {
-		if ( function_exists( 'the_custom_logo' ) && has_custom_logo() ) {
-			the_custom_logo();
+    if ( function_exists( 'the_custom_logo' ) && has_custom_logo() ) {
+      $logo = get_custom_logo();
+
+      echo $logo = is_home() ? '<h1>' . $logo . '</h1>' : $logo;
```
opportus: It makes perfect sense to wrap the logo in `<h1>` on home page. Especially that there is not category title to wrap multiple posts in.

Carefull the `<h1>` tag on custom logo has to be styled or wrapped in something. It actually takes the full col width.
It might also break something in header layouts of Storefront Designer.

single.php
opportus: Yes, better.

inc/storefront-template-functions.php
jameskoster: Is it definitely worth wrapping an image in a `h1`? 

Like you said this is gonna mess with the designer extension and probably all of the child themes as well. That's not ideal xD

jameskoster: If the h1 wrapper is preferred, let's also tweak `storefront_site_branding()` so that that `site-branding` wrapping div also wraps the logo. We can then use that class for all layout purposes. Then it should just be a case of removing the logo specific styles which is easy enough.

opportus: > If the h1 wrapper is preferred, let's just tweak storefront_site_branding() so that that site-branding wrapping div also wraps the logo. We can then use that class for all layout purposes. Then it should just be a case of removing the logo specific styles which is easy enough.

Yes, the `<h1>` is great on home page.
When it wraps the image, its content will be the `alt` attribute's value.

Seems a better idea to me also.
Since it's kinda CSS related, I let you deal with that one ?

opportus: @jameskoster 

> If the h1 wrapper is preferred, let's just tweak storefront_site_branding() so that that site-branding wrapping div also wraps the logo. We can then use that class for all layout purposes. Then it should just be a case of removing the logo specific styles which is easy enough.

Yes, the `<h1>` is great on home page.
When it wraps the image, its content will be the `alt` attribute's value.

Seems a better idea to me also.
Since it's kinda CSS related, I let you deal with that one ?

jameskoster: Yeah I'll handle all the styling if you tweak the markup.

404.php
```diff
@@ -22,44 +22,52 @@
 					<p><?php esc_html_e( 'Nothing was found at this location. Try searching, or check out the links below.', 'storefront' ); ?></p>
 
 					<?php
-					if ( is_woocommerce_activated() ) {
-						the_widget( 'WC_Widget_Product_Search' );
-					} else {
-						get_search_form();
-					}
+          echo '<section aria-label="Search">';
```
claudiosmweb: Need to fix this spaces and start use tabs

opportus: I do use tabs (2 indentation spaces). Any idea where it comes from ?

claudiosmweb: This is a WordPress project and WordPress coding standards recommend tabs...
I just add a .editorconfig file in this project, so now you only need to install in vim: https://github.com/editorconfig/editorconfig-vim#readme

opportus: I **DO** use tabs. 😕

inc/storefront-template-functions.php
opportus: @jameskoster 

See 1493485

I'll confirm when I'm done with everything.

opportus:
I think that's all.

We've got an SEO optimized Storefront !

jameskoster:
Woohoo! I'll check this all out soon and fix any spacing/tabbing issues.

Thanks so much @opportus, this is a great enhancement that I know our users will love.

opportus:
No problem, I'm glad having participated to it :)

inc/storefront-template-functions.php
```diff
@@ -158,19 +158,28 @@ function storefront_header_widget_region() {
...
-			<div class="site-branding">
-				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
-				<?php if ( '' != get_bloginfo( 'description' ) ) { ?>
+	function storefront_site_branding() { ?>
+    <div class="site-branding">
+      <?php
+      if ( function_exists( 'the_custom_logo' ) && has_custom_logo() ) {
+        $logo = get_custom_logo();
+
+        echo $logo = is_home() ? '<h1>' . $logo . '</h1>' : $logo;
```
opportus: Hi @jameskoster 

I see that you changed this to `is_front_page()` in the last update.

The problem with this new approach becomes obvious if we use a "static" one as front page like the WooCommerce Shop page... We don't want the logo as heading on this page but only the page title at the top level.

To make short, wrapping the logo inside `<h1>` is good only on home page (Posts page) since it doesn't have a heading itself.

jameskoster:
OK, will revert that.

inc/storefront-template-functions.php
jameskoster: OK, will revert that.

PR #442, opened 2016-07-28
Fix PHP 5.2 support

### LINKED ISSUE #441: Lost support for PHP 5.2
Since now are using closures, Storefront now supports only PHP 5.3 or later...

https://github.com/woothemes/storefront/blob/master/inc/class-storefront.php#L339-L341

https://github.com/woothemes/storefront/blob/master/inc/customizer/class-storefront-customizer.php#L74-L76

claudiosmweb:
Closes #441

### DIFF
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -336,11 +336,32 @@ if ( ! class_exists( 'Storefront' ) ) :
 			} else {
 				$structured_data = $structured_data + self::$structured_data[0];
 			}
-      array_walk_recursive( $structured_data, function ( &$value ) {
-        $value = sanitize_text_field( $value );
-      } );
+
+			$structured_data = $this->sanitize_structured_data( $structured_data );
 
 			echo '<script type="application/ld+json">' . wp_json_encode( $structured_data ) . '</script>';
+		}
+
+		/**
+		 * Sanitize structured data.
+		 *
+		 * @param  array $data
+		 * @return array
+		 */
+		public function sanitize_structured_data( $data ) {
+			$sanitized = array();
+
+			foreach ( $data as $key => $value ) {
+				if ( is_array( $value ) ) {
+					$sanitized_value = $this->sanitize_structured_data( $value );
+				} else {
+					$sanitized_value = sanitize_text_field( $value );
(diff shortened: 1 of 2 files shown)


inc/class-storefront.php
```diff
@@ -336,12 +336,33 @@ public function get_structured_data() {
 			} else {
 				$structured_data = $structured_data + self::$structured_data[0];
 			}
-      array_walk_recursive( $structured_data, function ( &$value ) {
-        $value = sanitize_text_field( $value );
-      } );
```
opportus: Oops, sorry.

inc/class-storefront.php
```diff
@@ -336,12 +336,33 @@ public function get_structured_data() {
...
+					$sanitized_value = $this->sanitize_structured_data( $value );
+				} else {
+					$sanitized_value = sanitize_text_field( $value );
+				}
+
+				$sanitized[ sanitize_text_field( $key ) ] = $sanitized_value;
+			}
+
+			return $sanitized;
+		}
```
opportus: Does `array_walk_recursive()` not recommended to use here ?

opportus: @claudiosmweb 

Does `array_walk_recursive()` not recommended to use here ?

opportus: @claudiosmweb 

Does `array_walk_recursive()` not good to use here ?

claudiosmweb: Does `array_walk_recursive()` sanitize the array keys too? :P

claudiosmweb: Only for values we can use only `wc_clean()`, but since I want to sanitize the keys too, I made this method.

opportus: I forgot that, I'll update this on my WooCommerce PR.

opportus: I forgot that I had to sanitize keys, thanks...

opportus: @claudiosmweb 
I forgot that I had to sanitize keys...
Also, you just learn me something pretty usefull with the method/function calling itself which by this way, recursively walk through the array, thanks...

PR #459, opened 2016-08-23
Initiate objects into an accessible global for access to hooks. Fixes #457

### LINKED ISSUE #457: Cannot `remove_action/filter` for hooks in `Storefront`
Since the Storefront object [is initiated without casting to a variable](https://github.com/woothemes/storefront/blob/master/inc/class-storefront.php#L401), or using a singleton, there is no way for me to unhook the `child_scripts` method (for example) from my child theme without re-creating the Storefront class myself (causing the `class_exists` check to return true). These hooks (as with most any hook in WordPress), should be accessible for removing.

jtsternberg:


jtsternberg:
This changes allows me access to the hooks within these classes, so I can do something like:

```php
add_action( 'after_setup_theme', function() {
	global $storefront;

	remove_action( 'wp_enqueue_scripts', array( $storefront->main, 'child_scripts' ), 30 );
	add_action( 'wp_enqueue_scripts', 'my_enqueue_child_scripts', 30 );

}, 5 );
```

jameskoster:
I think this should be fine. Gonna leave it to @tiagonoronha to review / merge though as I'm afk for a few days from this afternoon.

Couple of things; 

1. Would it be better to have a separate object per class? `$storefront`, `$storefront_woocommerce`, `$storefront_jetpack`, etc.

2. instead of having those null values, couldn't we do;

`'woocommerce' => class_exists( 'WooCommerce' ) ? require 'inc/woocommerce/class-storefront-woocommerce.php' : null,
'admin'       => is_admin() ? require 'inc/admin/class-storefront-admin.php' : null,`

Feels a little cleaner.

jtsternberg:
1. I like it all being under a one global 'namespace', `$storefront`, but if you prefer separate, it's an easy change.
2. Updated.

tiagonoronha:
Is there a specific reason we need to set `$storefront->woocommerce` and `$storefront->admin` on the initial declaration of the object?

I'd rather to this:

```
$storefront = (object) array(
	'version'     => $storefront_version,

	/**
	 * Initialize all the things.
	 */
	'main'        => require 'inc/class-storefront.php',
	'jetpack'     => require 'inc/jetpack/class-storefront-jetpack.php',
	'customizer'  => require 'inc/customizer/class-storefront-customizer.php'
);
```

and then

```
if ( is_woocommerce_activated() ) {
	$storefront->woocommerce = require 'inc/woocommerce/class-storefront-woocommerce.php';

	require 'inc/woocommerce/storefront-woocommerce-template-hooks.php';
	require 'inc/woocommerce/storefront-woocommerce-template-functions.php';
}

if ( is_admin() ) {
	$storefront->admin = require 'inc/admin/class-storefront-admin.php';
}
```

to keep things truly modular.

jtsternberg:
That's [how it was originally](https://github.com/woothemes/storefront/commit/ebf8fef8dc6ce601f6bb5c89ed91aa722a6c3f02), except I initiated the admin/woocommerce properties to `null`. I can certainly remove those null initiations if you prefer.

tiagonoronha:
Yup, lets go with that 👍  

And you can also move Jetpack to something like:

```
if ( class_exists( 'Jetpack' ) ) {
     $storefront->jetpack = require 'inc/jetpack/class-storefront-jetpack.php',
}
```

jtsternberg:
Ok, updated.

tiagonoronha:
Merged! Thanks, @jtsternberg :)

PR #477, opened 2016-09-26
Structured Data Update

opportus:


jameskoster:
Thanks @opportus !

jameskoster:
actually @opportus do you think we should leave the WooCommerce stuff in, but check the WooCommerce version before executing? Just to cover folks who update Storefront but don't update WooCommerce core for whatever reason.

opportus:
Hi @jameskoster 

With hindsight and gain of experience since I've included this functionality in Storefront, I now think that was a wrong choice in first place and that is not Storefront's job to do that.
But now it's implemented since a while so you may have a good point...
Also I've removed completely the function, but don't you think we should better deprecate it instead ?

As a side note, the `storefront_init_structured_data()` function generating schema.org for both items "Post" and "Page" is not an optimal approach. By separating this into 2 dedicated generator functions we would lower conditional checks.
But in this case, `storefront_structured_data` filter hook and the function itself are impossible to soft deprecate. This actual filter hook would be lost anyway, but that's also where it's leading about the  `storefront_woocommerce_structured_data` filter hook...
Do you think we should better leave it like that or refactor it optimally but making user's schema customizations lost ?

jameskoster:
@opportus hmm that's tricky. It's unlikely that anyone is interacting with that filter imo. But we can't say for sure. IMO;

* Add the WooCommerce version check wrapper to account for WooCommerce <2.7
* Leave `storefront_init_structured_data()` as it is for now
* Refactor (separate posts/pages) and remove the WooCommerce stuff altogether in 2.2.

What do you think?

opportus:
@jameskoster, so this means you plan to release Storefront v2.2 a while **after** WC v2.7 release, right ?
If so, that sounds all good to me.

I open a new PR as soon as possible.

jameskoster:
Yeah, Storefront 2.2 will mostly likely be after WooCommerce 2.7.

Issue #479, opened 2016-09-28
Header cart dropdown scrolling


jameskoster:
Currently the header cart contents dropdown scrolls as more products are added. This is to prevent a situation where the dropdown is so long the bottom of it goes off-screen. The scrolling can be distracting though. So let's only scroll if the height of the dropdown exceeds the window height.

PR #480, opened 2016-09-29
Header Cart Dropdown Scrolling

opportus:
Related to #479.

Actually, I'm not really experimented with JS, so take a close look... :P 
My main doubts are:

1. Does Storefront support Explorer 8 ? If so, I need to make it compatible...
2. Is the frontend behaviour OK ?
3. Is there a better approach ?

### DIFF
--- a/assets/js/woocommerce/header-cart.js
+++ b/assets/js/woocommerce/header-cart.js
@@ -11,18 +11,18 @@
 
 		cart.addEventListener( 'mouseover', function() {
 			var windowHeight  = window.outerHeight,
-				cartBottomPos = document.querySelector( '.site-header-cart .widget_shopping_cart_content' ).getBoundingClientRect().bottom + cart.offsetHeight,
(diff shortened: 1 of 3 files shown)


jameskoster:
This is a different approach to what I had in mind, but maybe I didn't explain that clearly in the original issue.

I think the header cart contents should only scroll if the dropdown is tall enough to make the bottom of it disappear off-screen.

What I was going to do was;

* Get the window height
* Get the site header height
* Get the cart dropdown height

Then, if the site header height + the cart dropdown height > the window height then add the scrolling behaviour.

opportus:
Hi @jameskoster 

It's what I've done... 😕

jameskoster:
Ah, it would help if I remembered to compile the css when I test, it looked like it was always scrolling. Derp xD

I didn't actually know about `getBoundingClientRect()`, definitely looks like a nice shortcut. Browse support is fine and it'll degrade fine.

I'll merge this in.

opportus:
Great !

PR #486, opened 2016-10-05
Change Footer Credit from WooThemes to WooCommerce

BFTrick:
Changing footer credit from WooThemes to WooCommerce


jameskoster:
Thanks!

PR #494, opened 2016-10-08
Footer Rows Implementation

### LINKED ISSUE #492: Footer Rows
There are many kind of widgets that one can wish to add in a footer, especially on ecommerces. In this regard, the actual footer layout is very limited.

So what about adding new **Top Footer** widget regions, which would bring necessary flexibility to the footer and would better suit to general uses anyway ?

This screenshot should describe well enough what I mean...

gsuite-footer

**EDIT**

A new flexible and extensible approach is preferable by making use of a new filter hook:

``` php
apply_filters( 'storefront_footer_widget_rows', $n );
```

After its implementation, we can get the exact same layout as on the screenshot above.

opportus:
Milestone 2.2.
Ready to merge.
Related to the issue #492.

Implements `apply_filters( 'storefront_footer_widget_rows', $n )` for giving extensivity and flexibility to the footer.

### DIFF
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -143,17 +143,17 @@ if ( ! class_exists( 'Storefront' ) ) :
 			);
 
 
-			$footer_widget_rows    = intval( apply_filters( 'storefront_footer_widget_rows', 2 ) );
-			$footer_widget_regions = intval( apply_filters( 'storefront_footer_widget_regions', 4 ) );
+			$rows    = intval( apply_filters( 'storefront_footer_widget_rows', 2 ) );
+			$regions = intval( apply_filters( 'storefront_footer_widget_regions', 4 ) );
 
-			for ( $row = 1; $row <= $footer_widget_rows; $row++ ) {
-				for ( $region = 1; $region <= $footer_widget_regions; $region++ ) {
+			for ( $row = 1; $row <= $rows; $row++ ) {
+				for ( $region = 1; $region <= $regions; $region++ ) {
 					$footer = sprintf( 'footer_%1$d_%2$d', $row, $region );
 
 					$sidebar_args[ $footer ] = array(
-						'name'        => sprintf( __( 'Footer Row%1$d Region%2$d', 'storefront' ), $row, $region ),
+						'name'        => sprintf( __( 'Footer Row-%1$d Region-%2$d', 'storefront' ), $row, $region ),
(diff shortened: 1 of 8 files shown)


inc/class-storefront.php
```diff
@@ -135,23 +135,27 @@ public function widgets_init() {
...
-					'id'          => sprintf( 'footer-%d', $i ),
-					'description' => sprintf( __( 'Widgetized Footer Region %d.', 'storefront' ), $i )
-				);
+			for ( $row = 1; $row <= $rows; $row++ ) {
+				for ( $region = 1; $region <= $regions; $region++ ) {
+					$footer = sprintf( 'footer_%1$d_%2$d', $row, $region );
+
+					$sidebar_args[ $footer ] = array(
+						'name'        => sprintf( __( 'Footer Row-%1$d Region-%2$d', 'storefront' ), $row, $region ),
+						'id'          => sprintf( 'footer-%1$d-%2$d', $row, $region ),
```
opportus: We may have a problem with widget ID's here - if it's a problem in first place.
After update, users will loose their footer settings though their widgets will be put on side in "Inactive Sidebars" sections.

If you consider it a problem, there is a simple workaround that is ready for commit.
Keep the old naming convention, eg:
- `'footer-1'`
- `'footer-2'`
- `'footer-3'`
- `'footer-4'`
- `'footer-5'`
- and so on...

... This way:

```php
'id' => sprintf( 'footer-%d', $region + $regions * ( $row - 1 ) );
```
With the same approach in the template function

inc/class-storefront.php
```diff
@@ -142,24 +142,28 @@ public function widgets_init() {
...
-
-			for ( $i = 1; $i <= intval( $footer_widget_regions ); $i++ ) {
-				$footer = sprintf( 'footer_%d', $i );
-
-				$sidebar_args[ $footer ] = array(
-					'name'        => sprintf( __( 'Footer %d', 'storefront' ), $i ),
-					'id'          => sprintf( 'footer-%d', $i ),
-					'description' => sprintf( __( 'Widgetized Footer Region %d.', 'storefront' ), $i )
-				);
+			$rows    = intval( apply_filters( 'storefront_footer_widget_rows', 2 ) );
```
opportus: @jameskoster, should the default still be 1 row (instead of 2 actually) ?

PR #495, opened 2016-10-08
Fixed PHP 5.2 compatibility

claudiosmweb:
Required some custom code, since there's no way to make `jetpack_the_site_logo()` return values, so I had to copy and paste the code.


jameskoster:
Thanks a lot @claudiosmweb :)

PR #504, opened 2016-11-08
update reference to function name in comment

jeffreysbrother:
The function name in the comment appears to be incorrect.
"storefront_single_post_after" is also referenced in single.php, but I'm not entirely sure what it's relationship to "storefront_single_post_bottom" might be.


jameskoster:
Almost certainly a copy/paste fail ;)

Thanks!

Issue #546, opened 2017-02-15
Rember Me Checkbox below Login button


Droidism:
When a returning customer tabs/clicks the Returning Customer messagebox, the radio checkmark `Remember me` is shown below the button. 

![screen shot 2017-02-15 at 18 34 03](https://cloud.githubusercontent.com/assets/10343479/22987153/1feeba70-f3ae-11e6-9bd1-a2b32ead1ce1.png)

I have experimented with a solution and in `woocommerce.scss`, right below `line 1072` I have added 

```
label.inline {
	display: inline !important;
	margin-left: ms(1);
}
```

This seems to solve this issue. Where possible I would prefer to loose the `!important`, but that would mean that `line 1072` should appear later in the cascade. A.t.m. `!important` is needed to override the declared style on `line 1419 - 1422` in `woocommerce.scss`.

Issue #544, opened 2017-02-15
Blog Navigation Numbering aligning to the left side instead of centre


Droidism:
Line  1291 in `_base.scss` ( `float: left` ) causes the Blog navigation to float to the, well, left side, instead of centre. Removing this line helps to solve the issue, but I have not checked if any other elements are affected by the removal of `float: left`. Let me know if it does and slap me on the wrist.

PR #550, opened 2017-02-16
Fixes for #544, #546 and enhancement of the sidebar

Droidism:
Still not clear if I am doing right, but let's go.


jameskoster:
Thanks again :)

PR #555, opened 2017-03-03
only load nux classes in the appropriate wp version and when in the a…

jameskoster:
…dmin or customizer


jameskoster:
@tiagonoronha we need to ensure this only loads for first time users. We shouldn't be loading anything after the user has used the stater content / guided tour etc. Wasn't sure how best to do that though. We can chat about it next week.

PR #557, opened 2017-03-09
Fixed some wrongly defined arguments for wp_enqueue_style & wp_register_script

zoliszabo:
Fixed missing/wrong arguments in inc/woocommerce/class-storefront-woocommerce.php


jameskoster:
Thanks a bunch! :)

PR #564, opened 2017-04-10
Fix .storefront-primary-navigation width in Opera

tomicakorac:
CSS for `.storefront-primary-navigation` was causing it to become extremely wide in Opera. This caused the horizontal scroll bar to appear needlessly. Setting `overflow: hidden;` to `.site-header` fixes the problem. `overflow-x: hidden;` would also be sufficient.

### DIFF
--- a/assets/sass/base/_layout.scss
+++ b/assets/sass/base/_layout.scss
@@ -17,7 +17,7 @@
 	.site-header {
 		padding-top: ms(5);
 		padding-bottom: 0;
-		overflow: hidden;
+		overflow-x: hidden;
 
 		.site-branding {
 			display: block;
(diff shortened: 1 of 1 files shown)


jameskoster:
I'm pretty sure `overflow: hidden;` on `.site-header` is going to hide dropdowns in your menu. `overflow-x` would be preferred.

tomicakorac:
@jameskoster it turns out you're right. Additionally, `overflow-x: hidden` doesn't work either due to this known issue:

http://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue/6433475#6433475

I'll close this pull request for now and re-open if I come up with a stable solution.

PR #566, opened 2017-04-19
Fix calls to deprecated method get_cart_url

pdewouters:


jameskoster:
Cheers!

claudiosanches:
@pdewouters @jameskoster 

A bug has been introduced here, there is no `WC()->cart->wc_get_cart_url()` method, the correct it's a function called `wc_get_cart_url()`.
Note that `WC()->cart->get_cart_url()` is soft deprecated and will not cause any issue or warning for now, but this Pull Request will result in a FATAL error, like:

```
PHP Fatal error:  Uncaught Error: Call to undefined method WC_Cart::wc_get_cart_url()
```

pdewouters:
oh I apologize!

jameskoster:
That's ok I fixed in https://github.com/woocommerce/storefront/commit/457fc1323d2a848dd71f506de1aa4fe81bf2a121

pdewouters:
thanks, sorry for giving you extra work :(

jameskoster:
No problem, really :)

PR #576, opened 2017-05-16
Translatable `aria-label`

andreaskian:
Improvement suggestion: remove hardcoded english texts causing unwanted result on multilingual site.

### DIFF
--- a/inc/storefront-template-functions.php
+++ b/inc/storefront-template-functions.php
@@ -571,7 +571,7 @@ if ( ! function_exists( 'storefront_product_categories' ) ) {
 			 */
 			if ( false !== strpos( $shortcode_content, 'product-category' ) ) {
 
-				echo '<section class="storefront-product-section storefront-product-categories" aria-label="Product Categories">';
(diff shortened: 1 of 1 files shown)


jameskoster:
Rather than removing it, why not make it translatable?

https://codex.wordpress.org/Function_Reference/esc_attr_e

andreaskian:
Hi James, 

I was simply not thinking of that – Sorry. I have tried to update my PR to use the `esc_html__()` function.

andreaskian:
Do I need to update `/languages/storefront.pot` with my new translatable strings?

jameskoster:
No problem. I think `esc_attr__()` might be the more appropriate function to use here though.

> Do I need to update /languages/storefront.pot with my new translatable strings?

No that's OK, our build tasks will take care of that when we release.

andreaskian:
I've updated PR with `esc_attr__()` instead

jameskoster:
Thank you! I'll merge this now.

PR #588, opened 2017-05-24
text color for alert on checkout

RistoNiinemets:
... alert that is shown if there are no available payment methods


jameskoster:
Thanks for the PR. It's not the color of the text that needs fixing here though, it's the background of the notice which should be blue.

The bottom padding needs fixing too. Would you like to update your PR?

RistoNiinemets:
I will, but I’m away for a week. 😊

jameskoster:
I fixed this in https://github.com/woocommerce/storefront/commit/1e132bdfe61d058121061427520cef0d78159d41

PR #593, opened 2017-05-28
amend support link in admin header

dougaitken:
Currently the "Support" link on `/wp-admin/themes.php?page=storefront-welcome` links to `https://support.woocommerce.com/` which is non-existent. Changing link to WordPress.org support forum page.


tiagonoronha:
👍

PR #612, opened 2017-06-25
hotfix header overlay-x

reatlat:
fix horisontal scroll on header part

### DIFF
--- a/assets/sass/base/_base.scss
+++ b/assets/sass/base/_base.scss
@@ -270,6 +270,8 @@ body {
 
 .site {
 	overflow-x: hidden;
+	position: relative;
+	display: initial;
 }
 
 .site-content,


reatlat:
image

jameskoster:
Which browser is this happening in? I thought we'd conquered this issue once and for all!

reatlat:
Hmm I use Chromium Version 59.0.3071.104 (Developer Build) (64-bit)
Maybe it happened only for Linux users...

jameskoster:
Interesting. And definitely latest Storefront version?

reatlat:
No it was old version 2.1.8,
I just update to version 2.2.3
and problem is almost same without background but with horizontal scroll
image

reatlat:
@jameskoster I just understand, this issue comes only then I resize a window screen.
I think it no big deal but sometimes users can resize a window and have a same issue...

mikejolley:
Latest Storefront no longer has this overlay so the fix won't be needed.

Tested on mac chrome, firefox, linux chromium.

Thanks

mikejolley:
Ah confirmed what this fixes on linux. Will cherry pick. Thanks.

Issue #615, opened 2017-07-03
Design conflict on Checkout page with demo store notice


iamdharmesh:
Design conflict happens on Checkout page between order review box and demo store notice.

When user change country review box's block overlay create conflict with demo store notice. 
Please check below screenshot for more information.

![storefront_checkout_block_overlap](https://user-images.githubusercontent.com/10613171/27801507-0e863632-603c-11e7-942e-1d09c487598a.png)

PR #616, opened 2017-07-03
Fixed issue #615 "Design conflict on Checkout page with demo store notice"

iamdharmesh:
Fixed issue #615 "Design conflict on Checkout page with demo store notice"


tiagonoronha:
Thanks 👍

PR #617, opened 2017-07-04
Add padding to shopping cart drop-down

mikeyarce:
Before: 
screen shot 2017-07-04 at 10 03 50 am

After:
screen shot 2017-07-04 at 10 05 25 am

### DIFF
--- a/assets/sass/woocommerce/woocommerce.scss
+++ b/assets/sass/woocommerce/woocommerce.scss
@@ -56,6 +56,7 @@
 
 	.widget_shopping_cart {
 		display: none;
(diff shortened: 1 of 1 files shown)


tiagonoronha:
@mikeyarce instead of adding that padding value manually, please use `ms(2)` instead. :)

```
.woocommerce-mini-cart__empty-message {
	margin: ms(2);
}
```

You let me know and I can do that for you. Thanks! 👍

mikeyarce:
@tiagonoronha @jameskoster good point!  Switched to using `ms` instead.

tiagonoronha:
🎉

PR #619, opened 2017-07-06
Fix bad line Issue in /inc/storefront-template-functions.php line 40

inetbiz:
Issue #621 https://github.com/woocommerce/storefront/blob/master/inc/storefront-template-functions.php#L40 had extra open bracket.


tiagonoronha:
https://github.com/woocommerce/storefront/issues/621#issuecomment-313382706

Issue #620, opened 2017-07-06
storefront functions [json+ld]


inetbiz:
Your image object needs to conform to http://schema.org/ImageObject and is missing some principle items. I would like to feature request the `caption`, `thumbnail` and `representativeOfPage` properties be inserted / amended.

inetbiz:
I'm not sure what to change in https://github.com/woocommerce/storefront/blob/master/inc/storefront-template-functions.php#L934 in order to get the `$attr` for **image description**. I think that should be used for http://schema.org/caption

tiagonoronha:
The structured date output is not present only in single pages/posts/products. It's also outputted to the homepage, archives, search results, etc.

For that reason, adding support for `representativeOfPage` is not that simple. Not all images are representative of the page.

I'll into this asap.

inetbiz:
@tiagonoronha ah. This is true. So append to string after logic filter?

inetbiz:
@tiagonoronha also review https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Foutput.jsbin.com%2Fpacodi for products and https://search.google.com/structured-data/testing-tool/u/0/?url=https://output.jsbin.com/vucuvex#url=https%3A%2F%2Foutput.jsbin.com%2Fvucuvex for category pages. courtesy of @jvandriel in our conversation at https://plus.google.com/+DenverProphitJr/posts/Mobiax1srhM for full context.

inetbiz:
Also this is how a category page should be written up. https://json-ld.org/playground/#/gist/c735f8a1c4f14523227e4064bfb61acb

inetbiz:
The main takeaway is there should never be more than one top level entity. A category page is a collection of like entities expressed in a nested entity of listitems that further have nested entities of product and offer.

A product page will be coming up this week. I added notes here since this function controls json+ld everywhere.

inetbiz:
@tiagonoronha guideline for a product page. https://json-ld.org/playground/#/gist/0453ffd28f753ec202501283b496947c

jvandriel:
> _"Also this is how a category page should be written up. https://json-ld.org/playground/#/gist/c735f8a1c4f14523227e4064bfb61acb"_

I'd not use the above mentioned template for category pages but https://output.jsbin.com/vucuvex instead.

tiagonoronha:
[...]

PR #623, opened 2017-07-06
Update storefront-template-functions.php

inetbiz:
Issue #620 updates issue with one attribute.

https://github.com/woocommerce/storefront/blob/master/inc/storefront-template-functions.php#L966
```
if ( $image ) {
				$json['image'] = array(
					'@type'  => 'ImageObject',
					'representativeOfPage'	=> 'true',
                                        'thumbnail' => //This needs to be nested with another ImageObject see http://schema.org/thumbnail
					'contentUrl'    => $image[0],
					'width'  => $image[1],
					'height' => $image[2],
				);
			}


tiagonoronha:
https://github.com/woocommerce/storefront/issues/620#issuecomment-313629032

PR #635, opened 2017-07-24
remove invalid parameter passed to remove_action

ashfame:
remove 4th parameter being passed to remove_action, it only accepts 3


tiagonoronha:
Thanks 👍

PR #638, opened 2017-08-03
Custom sidebar hook facility

manchumahara:
childtheme or plugin can hook to add extra sidebar without writing extra widget init hook and also this will work the same way other sidebar work

So, I suggest before sidebar register loop starts put a hook
$sidebar_args = apply_filters( 'storefront_sidebar_args', $sidebar_args );


tiagonoronha:
Thanks @manchumahara 👍

manchumahara:
welcome @tiagonoronha for accepting the PR

Issue #637, opened 2017-08-01
Unable to scroll to place order button with long payment method description


gedex:
I just set up a test store on WPCOM and use Storefront (2.2.4). Currently, there's a bug in Stripe where payment method description gets concatenated. However, with this long description, I'm not able to scroll to click the place order button. Here's the recording of it:

![storefront-issue](https://user-images.githubusercontent.com/78313/28812435-f967954a-76be-11e7-9711-d0f8b007c463.gif)

/cc @jameskoster as we had talk about this before

tiagonoronha:
@gedex what do I need to do to replicate this issue? 

I tried different methods and no mater how long the description is, scrolling still works.

gedex:
@tiagonoronha 

This is reproducible on my test site:

* Use incognito to open https://akeda-test-at-01.blog/
* Add one product on that shop page to the cart
* Go to the checkout page
* Scroll to the bottom of the page
* Choose **Credit Cart (Stripe)**
* Scroll up a bit

The steps at checkout page seem intermittent, so here's a way to reproduce it consistently:

![storefront-scroll-issue](https://user-images.githubusercontent.com/78313/28904477-6fa40884-7835-11e7-96e8-d45b6b1ad75b.gif)

I haven't tested this locally with `master` branch.

mikeyarce:
I wasn't able to reproduce this on my own site - I wonder if the height of the screen has to be a certain size for this to happen?

gedex:
I'm not sure. My window size: 1920px x 1173px.

mikeyarce:
I can reproduce this on your test site, when my browser is set to 80% size (probably because I'm on a 13'' Macbook pro and you have a 15'' macbook pro?

I can't reproduce this on my own site.

@jameskoster what are your thoughts on removing this feature entirely?  My reasoning is that it feels a bit unexpected when it's there - I have always preferred things not moving around.

gedex:
> and you have a 15'' macbook pro?

Correct. I'm using 15% MBP.

jameskoster:
> @jameskoster what are your thoughts on removing this feature entirely? My reasoning is that it feels a bit unexpected when it's there - I have always preferred things not moving around.

I don't know... I still believe having the purchase button permanently visible is a useful feature. Maybe we could simplify it so that only the purchase button sticks, rather than the entire order summary? That would reduce the scope for quirks like this dramatically.

[...]

PR #654, opened 2017-08-31
No jquery dependency on client-side code.

### LINKED ISSUE #631: jQuery-independent theme to increase performance
Hello, thanks for the awesome theme, it's indeed a great starting point!

I've been testing the theme with [LightHouse](https://chrome.google.com/webstore/detail/lighthouse/blipmdconlkpinefehnmjammfjpmpbjk?hl=en) and noticed that First Meaningful Paint can be improved of almost a whole second when jQuery is not loaded (I've modified `wp-content/script-loader.php` and removing jQuery and full jQuery UI scripts):

| with jQuery | without jQuery |
|----------|--------|
| screen shot 2017-07-19 at 5 31 06 pm | screen shot 2017-07-19 at 5 31 19 pm |

It would be great to see this theme being jQuery-agnostic and use native Web Platform features. 

After skimming quickly through the code, this looks like very much possible, and would be a very much welcomed win for our end-users which would get see the page refreshing a whole second faster! 🤑

WDYT? I can contribute through a PR, but before doing that I'd like to hear your opinion.

valdrinkoshi:
Fixes #631 
Remove jquery dependency from client side code.
This PR applies 2 fixes:
- replace jQuery APIs with native DOM APIs, some examples:
  - `jQuery(document).ready(...` => `document.addEventListener('DOMContentLoaded', ...`
  - `jQuery(element).on('click', ...` => `element.addEventListener('click', ...`
  - `jQuery(document).find(...` => `document.querySelectorAll(...`
- improve performance on `scroll/resize` events by debouncing callback to next animation frame

The admin client has been purposefully left untouched in this PR, since wordpress itself relies on jQuery.

The PR relies on `NodeList.forEach` to loop through elements, but that's not implemented in Internet Explorer, though a polyfill can be used in that case https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Polyfill - let me know if that's needed.

### DIFF
--- a/assets/js/homepage.js
+++ b/assets/js/homepage.js
@@ -16,6 +16,7 @@
 			return;
 		}
 		var siteMain = document.querySelector('.site-main');
+		var htmlDirValue = document.documentElement.getAttribute('dir');
 		var updateDimensions = function() {
 			if (updateDimensions._tick) {
 				cancelAnimationFrame(updateDimensions._tick);
@@ -24,7 +25,7 @@
 				updateDimensions._tick = null;
 				// Make the homepage content full width and centrally aligned.
 				homepageContent.style.width = window.innerWidth + 'px';
-				if (document.documentElement.getAttribute('dir') !== 'rtl') {
+				if (htmlDirValue !== 'rtl') {
 					homepageContent.style.marginLeft = -siteMain.getBoundingClientRect().left + 'px';
 				} else {
 					homepageContent.style.marginRight = -siteMain.getBoundingClientRect().left + 'px';
--- a/assets/js/homepage.js
+++ b/assets/js/homepage.js
@@ -3,74 +3,74 @@
  *
  * Handles behaviour of the homepage featured image
  */
-(function() {
+( function() {
 
 	/**
 	 * Set hero content dimensions / layout
 	 * Run adaptive backgrounds and set colors
 	 */
-	document.addEventListener('DOMContentLoaded', function() {
-		var homepageContent = document.querySelector('.page-template-template-homepage .type-page.has-post-thumbnail');
-		if (!homepageContent) {
+	document.addEventListener( 'DOMContentLoaded', function() {
+		var homepageContent = document.querySelector( '.page-template-template-homepage .type-page.has-post-thumbnail' );
+		if ( !homepageContent ) {
 			// Only apply layout to the homepage content component if it exists on the page
 			return;
 		}
-		var siteMain = document.querySelector('.site-main');
-		var htmlDirValue = document.documentElement.getAttribute('dir');
+		var siteMain = document.querySelector( '.site-main' );
+		var htmlDirValue = document.documentElement.getAttribute( 'dir' );
 		var updateDimensions = function() {
-			if (updateDimensions._tick) {
-				cancelAnimationFrame(updateDimensions._tick);
+			if ( updateDimensions._tick ) {
+				cancelAnimationFrame( updateDimensions._tick );
 			}
-			updateDimensions._tick = requestAnimationFrame(function() {
+			updateDimensions._tick = requestAnimationFrame( function() {
 				updateDimensions._tick = null;
 				// Make the homepage content full width and centrally aligned.
 				homepageContent.style.width = window.innerWidth + 'px';
-				if (htmlDirValue !== 'rtl') {
+				if ( htmlDirValue !== 'rtl' ) {
 					homepageContent.style.marginLeft = -siteMain.getBoundingClientRect().left + 'px';
 				} else {
 					homepageContent.style.marginRight = -siteMain.getBoundingClientRect().left + 'px';
 				}
-			});
+			} );
 		};
 		// On window resize, set hero content dimensions / layout.
-		window.addEventListener('resize', updateDimensions);
+		window.addEventListener( 'resize', updateDimensions );
 		updateDimensions();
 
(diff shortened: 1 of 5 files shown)


assets/js/homepage.js
```diff
@@ -4,43 +4,29 @@
...
 
 	/**
-	 * Set the hero component dimensions and positioning
+	 * Set hero content dimensions / layout
+	 * Run adaptive backgrounds and set colors
 	 */
-	function homepageContentDimensions() {
-		var windowWidth	    = jQuery( window ).width();
-		var offset          = jQuery( '.site-main' ).offset();
+	requestAnimationFrame(function() {
```
valdrinkoshi: Previously, the implementation would wait for the document to be ready. By using `requestAnimationFrame`, we  can assume that the first frame is about to be rendered, aka the parsing of the document is done.
This shouldn't be necessary, as the homepage script is injected at the end of the document (hence the DOM is already created and can be queried).

assets/js/homepage.js
```diff
@@ -62,22 +48,37 @@
...
 	/**
-	 * On window resize
-	 * Set hero content dimensions / layout
+	 * Set the hero component dimensions and positioning
 	 */
-	jQuery( window ).resize( function() {
-		homepageContentDimensions();
-	});
-} )();
+	function homepageContentDimensions() {
```
valdrinkoshi: We delay the update of the styles to the next frame, as this method might be called very frequently during window resize events.

assets/js/navigation.js
```diff
@@ -6,27 +6,29 @@
  * Finally adds a class required to reveal the search in the handheld footer bar.
  */
 ( function() {
-	// Add class to footer search when clicked
-	jQuery( window ).load( function() {
-		jQuery( '.storefront-handheld-footer-bar .search > a' ).click( function(e) {
-			jQuery( this ).parent().toggleClass( 'active' );
-			e.preventDefault();
+	// Add class to footer search when clicked.
+	requestAnimationFrame( function() {
```
valdrinkoshi: Similar to homepage.js, this file is injected at the end of the document, so DOM is already available to be queried. We wait a RAF, but that shouldn't be needed.

assets/js/navigation.js
```diff
@@ -52,54 +54,30 @@
...
-			toggleFocusClassTouchScreen();
-		}
-
-	// Add focus to cart dropdown
-	jQuery( window ).load( function() {
-		jQuery( '.site-header-cart' ).find( 'a' ).on( 'focus.storefront blur.storefront', function() {
-			jQuery( this ).parents().toggleClass( 'focus' );
+	// Add focus class to li and cart dropdown
+	requestAnimationFrame(function() {
+		document.querySelectorAll('.main-navigation a, .secondary-navigation a, .site-header-cart a').forEach(function(anchor) {
```
valdrinkoshi: I've merged the focus toggling in this method. Not clear what's the behavior we desire to have when on mobile, will need some assistance on this one.

inc/woocommerce/class-storefront-woocommerce.php
```diff
@@ -114,13 +114,16 @@ public function woocommerce_scripts() {
...
 		public function star_rating_script() {
-			if ( wp_script_is( 'jquery', 'done' ) && is_product() ) {
+			if ( is_product() ) {
 		?>
 			<script type="text/javascript">
-				jQuery( function( $ ) {
-					$( 'body' ).on( 'click', '#respond p.stars a', function() {
-						var $container = $( this ).closest( '.stars' );
-						$container.addClass( 'selected' );
+				requestAnimationFrame(function() {
```
valdrinkoshi: Wasn't exactly sure where this script will be injected, so I've implemented it by assuming `p.stars` are already rendered by the next frame time (aka, no other stars elements are created dynamically at some later point)

claudiulodro:
This looks great. Everything works properly and I like the RAF improvements.

Couple notes:
1. It's going to be a little bit of a pain in the butt to go through and do this after everything's written, but [spacing in the JS needs to follow the WP style guidelines](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#examples-of-good-spacing). That's mostly going to be adding a space in function calls. For example, `(foo,bar) -> ( foo, bar )`

2. On Safari the checkout order review box sticks to the screen too early. This is probably because of a browser difference in how Safari calculates box heights or window heights. It's not a problem in other browsers (haven't tested Edge, though).
screen shot 2018-01-05 at 1 07 58 pm

@jameskoster @tiagonoronha How do you come up with the version number for the enqueued scripts (e.g. `'20120206'`)? These JS files will need a new version number.

tiagonoronha:
>On Safari the checkout order review box sticks to the screen too early. This is probably because of a browser difference in how Safari calculates box heights or window heights. It's not a problem in other browsers (haven't tested Edge, though).

We are removing that script very soon. See #637.

>How do you come up with the version number for the enqueued scripts (e.g. '20120206')? These JS files will need a new version number.

I don't think there's a particular logic behind that number. You can change that you use the version number of Storefront.

valdrinkoshi:
@claudiulodro thanks for the feedback. I will update the js space styling and give a look at the safari bug, just for fun 👌

valdrinkoshi:
@claudiulodro PTAL at the last 5 commits:
- fixed js space formatting
- now using storefront version for the js files
- fixed sticky positioning by simplifying the logic (it's enough to check for `getBoundingClientRect().top <= 0`, see last commit)

claudiulodro:
Excellent. Everything looks good and works well now. Thanks!

PR #658, opened 2017-09-11
Add Support for Memberships Members Area

bekarice:
The upcoming Memberships 1.9 release changes the account area to improve the members' area, which shows perks and restricted content. It adds a new account navigation item for Memberships: http://cloud.skyver.ge/1h3V462B0E0G  (moving member data here from the Dashboard)

This link then either (a) takes the member directly into the members' area if they have one membership, or (b) shows the "My Memberships" table if there are multiple memberships so the member can choose which area to view.

This PR adds support for showing an icon with the Memberships account navigation item, along with icons for the members area navigation itself: http://cloud.skyver.ge/141R3q213p14

(Memberships basically replaces the default account navigation tabs while in the members area with its own tabs.)

~~Note that, as the endpoint for this navigation item _could_ result in linking directly to the first section in the members area, the CSS class for it can change, which is why there's a wildcard CSS rule for it.~~ UPDATE: found a way to filter this class so it's consistent for membership count == 1 and > 1 😸


bekarice:
cc @jameskoster

jameskoster:
Thanks @bekarice !

PR #664, opened 2017-09-28
add  wp_kses ( __(  )  );

Umangvaghela:
use  wp_kses ( __(  )  ); to display html

### DIFF
--- a/inc/storefront-template-functions.php
+++ b/inc/storefront-template-functions.php
@@ -187,7 +187,7 @@ if ( ! function_exists( 'storefront_site_title_or_logo' ) ) {
 			$logo_id = get_theme_mod( 'custom_logo' ); // Check for WP 4.5 Site Logo
 			$logo_id = $logo_id ? $logo_id : $logo['id']; // Use WP Core logo if present, otherwise use Jetpack's.
 			$size    = site_logo()->theme_size();
-			$html    = sprintf( '<a href="%1$s" class="site-logo-link" rel="home" itemprop="url">%2$s</a>',
+			$html    = sprintf( wp_kses ( '<a href="%1$s" class="site-logo-link" rel="home" itemprop="url">%2$s</a>',
 				esc_url( home_url( '/' ) ),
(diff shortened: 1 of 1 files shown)


inc/storefront-template-functions.php
```diff
@@ -366,7 +366,7 @@ function storefront_post_header() {
 				storefront_posted_on();
 			}
 
-			the_title( sprintf( '<h2 class="alpha entry-title"><a href="%s" rel="bookmark">', esc_url( get_permalink() ) ), '</a></h2>' );
+			the_title( sprintf(  wp_kses ( __( '<h2 class="alpha entry-title"><a href="%s" rel="bookmark">', esc_url( get_permalink() ) ), '</a></h2>' ) ) );
```
claudiosanches: There is no need for `__()`, since it's just pain HTML here, there is no text do translate.

inc/storefront-template-functions.php
```diff
@@ -187,7 +187,7 @@ function storefront_site_title_or_logo( $echo = true ) {
 			$logo_id = get_theme_mod( 'custom_logo' ); // Check for WP 4.5 Site Logo
 			$logo_id = $logo_id ? $logo_id : $logo['id']; // Use WP Core logo if present, otherwise use Jetpack's.
 			$size    = site_logo()->theme_size();
-			$html    = sprintf( '<a href="%1$s" class="site-logo-link" rel="home" itemprop="url">%2$s</a>',
+			$html    = sprintf( wp_kses ( '<a href="%1$s" class="site-logo-link" rel="home" itemprop="url">%2$s</a>',
```
claudiosanches: Already escapes the image URL.

claudiosanches:
First off, thanks for sending in your contribution. I have some feedback:

- All this places already escape URLs
- `__()` should be used to translatable strings, and not for HTML
- None of the places applied requires any extra escaping for HTML, since all are hard coded, we should use `wp_kses()` when is some data submitted by an user.

That's why I'm closing.
Thanks again.

PR #674, opened 2017-11-04
Tweak product grid to align Add to cart buttons

thomasschmit:
Currently on a product grid, when the title of a product takes several lines, it creates a gap between the different *Add to cart* buttons. These few declarations tweak that glitch simply, using flexbox.

Here is a before/after comparison:

tweak_product-grid

### DIFF
--- a/assets/sass/woocommerce/woocommerce.scss
+++ b/assets/sass/woocommerce/woocommerce.scss
@@ -278,12 +278,18 @@ table.shop_table_responsive {
  * Products
  */
 ul.products {
+	display: flex;
+	flex-wrap: wrap;
 	margin-left: 0;
 	margin-bottom: 0;
 	clear: both;
 	@include clearfix;
 
 	li.product {
+		align-items: center;
+  		display: flex;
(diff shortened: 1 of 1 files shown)


jameskoster:
I'm torn on this. I know the mis-alignment bugs some folks, but personally I find the inconsistent spacing between the prices / buttons really awkward as well. I'm not sure which is the worse offender.

@tiagonoronha what do you think?

tiagonoronha:
I'm also torn on this. 🤔 

Well, there have been more times where I wanted the buttons to be aligned than the opposite.

jameskoster:
Make it so

tiagonoronha:
@jameskoster Yeah, not sure: https://cloudup.com/cDKtBEy73Lo

jameskoster:
haha, yeah that looks awful. Using that evidence I think we should leave it as it is. This is a fairly simple tweak for people who feel passionately about it.

tiagonoronha:
@thomasschmit thanks for the PR, but for the reasons above, we are not going to accept it.

PR #690, opened 2017-11-22
remove need to install grunt globally

valdrinkoshi:
Use instead the grunt version installed by this package


mikejolley:
Works for me.

PR #700, opened 2018-01-02
WooCommerce 3.3 columns & rows

### LINKED ISSUE #680: Define WooCommerce min/max columns & rows
Once 3.3 lands we won't need to specify the exact number of columns / rows products are displayed in as core will offer an option to manage this.

We just need to define the min / max columns and rows.

See https://github.com/woocommerce/woocommerce/blob/1d93d8f577f3c7aabe632db90747bf07cd1ae86a/includes/admin/settings/class-wc-settings-products.php#L154-L169

We can also wrap our `loop_shop_columns` filter inside a version check and only filter for WC versions older than 3.3.

tiagonoronha:
Adds support for WooCommerce 3.3 default, min, and max columns and rows and ensures backwards compatibility with older versions.

Closes #680.

### DIFF
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -120,8 +120,7 @@ if ( ! class_exists( 'Storefront' ) ) :
 					'default_rows'    => 4,
 					'min_columns'     => 1,
 					'max_columns'     => 6,
-					'min_rows'        => 1,
-					'max_rows'        => ''
+					'min_rows'        => 1
 				)
 			) ) );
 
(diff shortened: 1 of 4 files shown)


tiagonoronha:
@jameskoster please test again :)

mikejolley:
@tiagonoronha same for me. Changing rows has no effect on the view.

tiagonoronha:
@mikejolley definitively works for me now: https://cloudup.com/c84h5vNBgMG

Please double check that you've the most recent commits for this branch.

tiagonoronha:
Found a different issue while testing. Now you definitively need to update. :)

PR #718, opened 2018-01-05
Price filter slider styling

tiagonoronha:
Old

screen shot 2018-01-05 at 18 06 59

New

screen shot 2018-01-05 at 18 06 16


mikejolley:
Looks good now 👍

PR #749, opened 2018-01-26
Prevent image bleed from next image in the gallery

tiagonoronha:
Reported here: https://github.com/woocommerce/homestore/issues/35

### DIFF
--- a/assets/sass/woocommerce/woocommerce.scss
+++ b/assets/sass/woocommerce/woocommerce.scss
@@ -406,7 +406,7 @@ ul.products {
 			.woocommerce-product-gallery__wrapper {
 				.woocommerce-product-gallery__image {
 					&:not(:only-child):not(.flex-active-slide) {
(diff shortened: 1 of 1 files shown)


tiagonoronha:
This needs more work, the transition is ugly with the initial fix.

tiagonoronha:
This issue is caused by Flexslider not correctly calculating the size of the container. We use percentages, and not a fixed value. In some specific cases, the calculation is off by about ~1px.

I tested and the fix doesn't seem to cause any other issues, I believe it is safe to merge.

mikejolley:
Looks ok to me 👍

dr1ftk1ng:
In my case i had to manually remove this fix, as it was causing the bleed.
I am running Woocommerce 3.3.3 with Storefront 2.2.8
I am not quite sure, but maybe the current Woocommerce version is using a fixed Flexslider version.

Issue #683, opened 2017-11-15
Cart table overflowing container


mikeyarce:
Just before the mobile breakpoint, the `table.cart` overlfows into the sidebar:
![screen shot 2017-11-15 at 11 51 44 am](https://user-images.githubusercontent.com/3220162/32857071-8cd8ff3c-c9fb-11e7-8855-0b0c25c747db.png)

Possible solution is just to apply a `table-layout: fixed;` to the `table.cart` but then that makes each row an equal size which is not ideal:

![screen shot 2017-11-15 at 11 53 46 am](https://user-images.githubusercontent.com/3220162/32857104-ad9c71ea-c9fb-11e7-9ca8-ef16079e9bfb.png)

We could also set widths for each row, which would solve that problem.

Alternatively, we could break the sidebar down earlier in the screen width.  Or, we could show the mobile cart earlier by changing the smallscreen-css breakpoint.

Thoughts @jameskoster @tiagonoronha ?

jameskoster:
Hmm, I think we should just decrease the table padding a bit when;

1. The cart page is not using the full width template
2. Viewport is larger than the mobile switch
3. Viewport is small enough to cause this awkward display

mcsmithers:
I can fix this.  I am assuming it's in the 'woocommerce.css' or the `woocoomerce-smallscreen.css`, right?

mcsmithers:
I think I found a fix to this.  If you use Storefront from WooCommerce and use it with out-of-the -box themes such as TwentySixteen or TwentyFifteen, the CSS rules conflict with each other.  I adjusted the storefront to have a little more padding on the right sidebar so things look tidier though when combined with other themes, you may see some overlap.

PR #750, opened 2018-01-29
Overlap with side widget fixed

mcsmithers:
**Problem:**  CSS in cart overlaps with right-sidebar widget as shown in issue [683](https://github.com/woocommerce/storefront/issues/683).  This is right before the mobile breakpoint by only a couple pixels.

Screenshot
image

**Solution**:  Modified `woocommerce.scss` to give a slight amount more room on the right side so that there's not overlap with the right sidebar widget.  

Result Screenshot:
image

**Caveat:** 
CSS fixed to give the table of the cart more room so that the side widget doesn't overlap.  Please note that other themes may overwrite this in their CSS.  This is especially the case with TwentySixteen and TwentyFifteen so far.

### DIFF
--- a/assets/sass/woocommerce/woocommerce.scss
+++ b/assets/sass/woocommerce/woocommerce.scss
@@ -56,6 +56,7 @@
 
 	.widget_shopping_cart {
 		display: none;
+		padding-right: 15px;
 
 		.woocommerce-mini-cart__empty-message {
 			margin: ms(2);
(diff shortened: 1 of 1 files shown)


tiagonoronha:
Hi @mcsmithers. Thank you for the PR. Unfortunately, this is not the type of fix we are look for this issue.

Please read this comment: https://github.com/woocommerce/storefront/issues/683#issuecomment-344939327

Feel free to send a new PR.

PR #767, opened 2018-02-09
Reduce padding on cart table to prevent sidebar from overflowing container

### LINKED ISSUE #683: Cart table overflowing container

tiagonoronha:
Closes #683.

### DIFF
--- a/assets/sass/woocommerce/woocommerce.scss
+++ b/assets/sass/woocommerce/woocommerce.scss
@@ -2791,7 +2791,7 @@ dl.variation {
 	}
 }
 
-@media (min-width: $handheld) and (max-width: 900px) {
+@media (min-width: $desktop) and (max-width: 900px) {
 	body {
 		&:not(.page-template-template-fullwidth-php) {
 			table.cart {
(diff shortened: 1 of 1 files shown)


claudiulodro:
Looks good. This fixes the original issue. There is slightly more padding from 568px-900px breakpoint on items.

Before:
screen shot 2018-02-12 at 10 21 58 am

After:
screen shot 2018-02-12 at 10 22 06 am

I think it looks fine, just a heads-up. Go ahead and merge this if you're OK with that.

tiagonoronha:
@claudiulodro great find! I updated the breakpoint to be the Desktop (768px) instead.

claudiulodro:
Looks good. That's solved the issue.

PR #792, opened 2018-03-14
Single Product Pagination

### LINKED ISSUE #766: Add Storefront Product Pagination to core
https://wordpress.org/plugins/storefront-product-pagination/

This extension is used by about 30% of all Storefront active installations.

tiagonoronha:
To test:

1. Checkout this branch
2. Go to Single Product page
3. You should see:

screen shot 2018-03-14 at 17 59 07

On hover:

screen shot 2018-03-14 at 17 59 51

Closes #766.


jameskoster:
@tiagonoronha a couple of thoughts;

* What happens if they have the plugin installed as well? We should account for that as it will obviously affect anyone who currently has the plugin active.
* Should we consider adding an option to disable the feature?

tiagonoronha:
@jameskoster 

> What happens if they have the plugin installed as well? We should account for that as it will obviously affect anyone who currently has the plugin active.

https://github.com/woocommerce/storefront/pull/792/commits/d04997e4f78c792f4160a7eff25d0243a91c838f 

> Should we consider adding an option to disable the feature?

It's worth thinking about it, how do you feel about adding more options to the Customizer? Alternative we can provide documentation on how to remove it with `remove_action`, although not as friendly.

mikejolley:
@tiagonoronha This works. I agree with option:

> It's worth thinking about it, how do you feel about adding more options to the Customizer? Alternative we can provide documentation on how to remove it with remove_action, although not as friendly.

Customizer +1

tiagonoronha:
Added toggle to the Customizer:

screen shot 2018-04-17 at 19 23 00

sandrodz:
https://github.com/woocommerce/storefront/issues/864

PR #830, opened 2018-04-17
Improve semantic HTML for page

seb86:
Change page wrapper from **div** to **article** for single pages to improve semantic HTML.


tiagonoronha:
Thanks! :)

PR #836, opened 2018-04-26
[2.3] Improve sticky button behavior with non-core product types

franticpsyx:
Because the sticky button element is now on by default, it might make sense to ensure it works reliably with non-core product types. I did some tests and found a couple issues -- hopefully this PR addresses them well. 

1. The sticky button text is not generic enough. A more suitable method to populate it might be `add_to_cart_text`. My feeling is it works a bit better than the existing conditional logic that uses `single_add_to_cart_text`. It produces the expected result with all product types I tested.
2. Allow customizing the "trigger" element. Some product types may include options to display add-to-cart form content after the 'entry-summary' element. In such cases, the sticky-button behavior should be customizable (ideally) to work reliably.

### DIFF
--- a/assets/js/sticky-add-to-cart.js
+++ b/assets/js/sticky-add-to-cart.js
@@ -1,3 +1,4 @@
+/*global storefront_sticky_add_to_cart_params */
 ( function() {
 	document.addEventListener( 'DOMContentLoaded', function() {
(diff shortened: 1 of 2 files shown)


tiagonoronha:
@franticpsyx I love this! ❤️ 

Thanks so much!

PR #906, opened 2018-06-03
Fix composited bundles layout issues

### LINKED ISSUE #905: [2.3] Composited bundles layout issues
Composited product image styles appear to be taking precedence over bundle item image styles, resulting in this:

image 2018-06-03 at 11 24 54 p m

Not sure which commit might have triggered this but I'm pretty certain this is not an old issue, as we've been seeing reports of this in support only since recently.

PR with fix coming soon.

franticpsyx:
Closes #905


tiagonoronha:
👍

PR #912, opened 2018-06-08
WooCommerce integration code organization

tiagonoronha:
This PR makes changes to the structure of the theme to allow for better separation between the theme and the WooCommerce integration. This way, WooCommerce related code will only load if the plugin is actually enabled.

Other changes:

* Removes deprecated functions
* Replaces 'WooThemes' with 'Automattic'


tiagonoronha:
Removed the `@author` tag. Merging :)

PR #926, opened 2018-06-27
Add taxonomy and extra filters to product pagination

### LINKED ISSUE #924: Product Pagination "same category" bug
Hi, 

product pagination does not work when `storefront_single_product_pagination_same_category` filter returns true. get_previous_post and get_next_post default taxonomy is _category_. We need _product_cat_.

So ...

`$previous_product = get_previous_post( $same_category );` 

... should be ...

`$previous_product = get_previous_post( $same_category, '', 'product_cat' );`

tiagonoronha:
We include a filter in product pagination to allow for fetching only posts from the same category. It's currently not working as expected because it requires a third parameter, the taxonomy, to be set.

This PR also includes two additional new filters:

* `storefront_single_product_pagination_excluded_terms`
* `storefront_single_product_pagination_taxonomy `

Closes #924.

### DIFF
--- a/inc/woocommerce/storefront-woocommerce-template-functions.php
+++ b/inc/woocommerce/storefront-woocommerce-template-functions.php
@@ -709,11 +709,13 @@ if ( ! function_exists( 'storefront_single_product_pagination' ) ) {
 		}
 
 		// Show only products in the same category?
-		$same_category = apply_filters( 'storefront_single_product_pagination_same_category', false );
+		$in_same_term   = apply_filters( 'storefront_single_product_pagination_same_category', false );
+		$excluded_terms = apply_filters( 'storefront_single_product_pagination_excluded_terms', '' );
+		$taxonomy       = apply_filters( 'storefront_single_product_pagination_taxonomy', 'product_cat' );
 
 		// Get previous and next products.
-		$previous_product = get_previous_post( $same_category );
-		$next_product     = get_next_post( $same_category );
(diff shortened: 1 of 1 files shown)


SimplyBoho:
Does not work correctly if the product is in more than one category.

tiagonoronha:
@SimplyBoho what do you mean? Any examples?

SimplyBoho:
Here is my example:
I use Storefront theme and sell baby onesies. I have a design that is "boho style" and is for girls only. So I put it in "Girl Designs" AND "Boho Designs". When i'm viewing the "Boho Designs" category, and click the pagination for this particular girls boho design, the pagination switches to "Girl Designs". I really want it to stay in "Boho Designs" and show only those items, since they include boy and girl designs.

Does that make sense?

SimplyBoho:
You are correct, I didn't even notice that! My breadcrumbs do the same
thing. I'll have to regthink all this then :)
(I'm still building my site so I don't have much listed)

tiagonoronha:
@rodrigoprimo this was missing the new variables in for the parameters in some of the functions. See: https://github.com/woocommerce/storefront/pull/926/commits/cce0b284763d3b91f933028c39cceed4f576a6f9

peterfabian:
I like the change, it's more configurable now and solves the problem. 

One thing that is perhaps suboptimal is the use of `get_previous_post` and `get_next_post` functions. If two products are created at the same time (which I assume is possible during e.g. import from csv), then some of them are skipped when going back and forth between products.

PR #937, opened 2018-07-12
Prevent select from overflowing container

tiagonoronha:
Fixes overflow issue noticed in https://github.com/woocommerce/woocommerce/issues/20756#issue-340209496

* Sets global select `max-width` to `100%`
* Sets variations select `max-width` to `75%` so that the "clear" button is able to line up side by side with the dropdown.

Before:

img_5740

After:

img_5739

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -44,6 +44,7 @@ textarea {
 select {
 	color: initial;
 	font-family: $base-font;
+	max-width: 100%;
 }
 
 h1,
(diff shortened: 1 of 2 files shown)


peterfabian:
All desktop browsers and all non iOS-Safari ones seem to not wrap, but cut off the text. Therefore, I would maybe split it at maybe 80/20 to get a bit more of the text?
Otherwise works fine. 

select-width

tiagonoronha:
I'm happy with 70% as it leaves room for the "Clear" button to expand on different languages.

PR #938, opened 2018-07-17
Add `storefront_after_site` action

JoryHogeveen:
There is a `storefront_before_site` action but no `storefront_after_site`. This PR adds this action to footer.php.

I've placed it before `wp_footer()` since that hook is usually for scripts etc.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -34,6 +34,8 @@
 
 </div><!-- #page -->
 
+<?php do_action( 'storefront_after_site' ); ?>
+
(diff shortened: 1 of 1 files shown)


claudiosanches:
`wp_footer()` just include a `wp_footer` hook, so that's why there is no `storefront_after_site`, since you can already use `wp_footer`. Note that `storefront_before_site` comes after `<body>` where by default here is no other hook.

Thanks for your contribution.

Issue #953, opened 2018-08-30
Storefront WC options should display only if WC is activated


GlennMartin1:
In Storefront 2.3.0, **Sticky Add-To-Cart** and **Product Pagination** was added for WooCommerce products.

**BUG:** These options display in the Customizer even if WooCommerce is not installed/activated.

**Expected:** these options should only appear in the Customizer if WooCommerce is active.

tiagonoronha:
This is coming in Storefront 2.4. If you download from master, you'll see this already implemented.

https://github.com/woocommerce/storefront/pull/954#issuecomment-418751575

Changes were done here: https://github.com/woocommerce/storefront/pull/912

Closing :)

PR #954, opened 2018-08-30
Added theme support to the Customizer section

mahdiyazdani:
Makes sure WooCommerce is activated before adding a customize section
Related to #953


tiagonoronha:
Hi @mahdiyazdani,

Storefront 2.4, coming soon, moves all WooCommerce related code to separate files that are only loaded when the plugin is active.

This code is already finished and in master. The file you edited `class-storefront-woocommerce-customizer.php` is already only loaded when WooCommerce is active.

I understand this is a bit confusing when comparing Storefront 2.3 with the code in master.

For this reason, this PR is not necessary.

Thank you for your contribution.

GlennMartin1:
> Storefront 2.4, coming soon

approx ETA?

PR #959, opened 2018-09-06
Fix clearing of cart item rows on handheld screens

### LINKED ISSUE #958: Cart item rows not cleared properly (handheld/mobile screens)
Cart item `td` elements are nicely styled using `display:block` on mobile screen sizes -- however, these rows aren't always cleared properly. Here's an example:

image 2018-09-06 at 11 23 35 p m

Might be a good idea to add `@include clearfix` somewhere? PR coming.

franticpsyx:
Closes #958


tiagonoronha:
👍

PR #965, opened 2018-09-18
Touch devices dropdown support

### LINKED ISSUE #962: Dropdown menu does not open correctly on a tablet
If a drop down menu with the top-level menu item as a link is tapped in tablet view, the drop down menu does not open. Instead, the top link opens. 

Ticket here: 1406543-zen

tiagonoronha:
Adds touch support for navigation dropdowns when using an handheld device such as an iPad.

How to test:

1. Create a menu with dropdowns and assign it to the "Primary navigation" area.

2. Use an iPad or switch Chrome to mobile mode and select the iPad Pro from the list of devices (if using Chrome, reload the page after switching to the iPad). 

3. Tap on the drop down to open it. The first tap should open the dropdown. A second tap on the same item should navigate away to the correct page.

Closes #962.

### DIFF
--- a/assets/js/navigation.js
+++ b/assets/js/navigation.js
@@ -130,9 +130,7 @@
 						el.classList.remove( 'focus' );
 
 						// Remove blocked class, if it exists.
-						if ( el.firstChild && el.firstChild.classList ) {
-							el.firstChild.classList.remove( 'blocked' );
-						}
+						el.firstChild.classList.remove( 'blocked' );
 					}
 				} );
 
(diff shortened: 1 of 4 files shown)


assets/js/navigation.js
```diff
@@ -135,6 +147,32 @@
...
+						if ( ( 'cart-contents' === anchor.className && anchor.parentNode.nextElementSibling ) || anchor.nextElementSibling ) {
+							anchor.classList.add( 'blocked' );
+							aEvent.preventDefault();
+						}
+					} );
+				} );
+			} );
+
+			// Ensure the dropdowns close when user taps outside the site header
+			[].forEach.call( document.querySelectorAll( '.site-content, .header-widget-region, .site-footer' ), function( element ) {
```
peterfabian: The dropdown does not close if the user clicks in the header area, e.g. to search products via the input, etc. It's not critical, but would be nice if that behaved in the same way as when clicking on somewhere else on the page.

tiagonoronha:
@peterfabian It's really trick to allow for clicks inside the header area because I would need to basically listen to events on all elements, but I made a change that target a few elements in the Header so, for example, clicking on the search input, will close the dropdown.

PR #969, opened 2018-09-27
Support CP 3.14 and PB 5.8 (releases imminent)

franticpsyx:
This PR:

* Introduces new styles for Bundle /Composite-type single-product pages. Both Composite Products v3.14 and Product Bundles v5.8 include updates to the single-product page styles that need to be supported by Storefront. The change set here is backwards compatible with previous PB/CP versions.
* Adds support for some advanced theme features in CP/PB which improve the way Bundle/Composite child items are displayed on handheld screens in the cart (older PB/CP versions aren't affected):

image 2018-09-27 at 9 42 30 a m

@tiagonoronha Both releases are imminent (scheduled for Oct 15). It would be awesome if we could include this one in a Storefront patch/fix release before that date!

PS: I have included a few `@since` tags in a couple places, which reference `2.4.0` as I wasn't sure what the next Storefront version would be. Lmk if you need an extra commit from me to make those accurate.

### DIFF
--- a/assets/css/woocommerce/extensions/bundles.scss
+++ b/assets/css/woocommerce/extensions/bundles.scss
@@ -1,76 +1,187 @@
 /**
- * WooCommerce AJAX Layered Nav
+ * WooCommerce Product Bundles
  */
 
 /**
  * Imports
  */
+
 @import 'bourbon';
 @import '../../sass/utils/variables';
 @import '../../sass/utils/mixins';
 @import '../../../../node_modules/susy/sass/susy';
 @import '../../sass/vendors/modular-scale';
 
+/**
+ * Base
+ */
+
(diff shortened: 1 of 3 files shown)


tiagonoronha:
@franticpsyx we have a release planned for the second week of October. That will be version 2.3.4.

Please update this PR, and instead of doing it against master, please use the release/2.3 branch. We will then later merge your commits to master.

franticpsyx:
Thanks @tiagonoronha -- the timing sounds awesome.

I'll be closing this one out and open a new one as I see there's quite a lot that's different in 2.4.

franticpsyx:
Opened https://github.com/woocommerce/storefront/pull/970

PR #986, opened 2018-11-01
Add version suffix to loaded css integration files (2.3)

### LINKED ISSUE #985: Integration css files not suffixed with version
### Description ###

The active version of Storefront is not appended to the filename when loading integration css files. Many browsers out there are using the local (cached) versions of these files.

Got quite a few reports from our customers about this (bundled cart items not indented). PR coming :)

franticpsyx:
closes #985


tiagonoronha:
No need to submit a separate PR for this. The file structure is the same so it is quite easy to cherry pick your other commit from master.

Thanks! :)

PR #989, opened 2018-11-02
Action hook right before closing <body> tag

lucspe:
Hi, I would like to submit my request to add an action hook right before the closing <body> tag.

That is, essentially, the counterpart to 'storefront_before_site' and would be useful to add custom code or small JS snippets at the very end.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -36,5 +36,7 @@
 
 <?php wp_footer(); ?>
 
+<?php do_action( 'storefront_after_site' ); ?>
+
 </body>
 </html>


lkraav:
What's wrong with `wp_enqueue_script()` with `$in_footer = true` parameter?

For PHP, there's are a metric ton of core actions you can run near the end body tag output. Not sure what exactly a real use case is for targeting the closing body tag very accurately?

lucspe:
Nothing wrong with wp_enqueue_script with $in_footer = true, if you want to add an external JS file.

As for core actions, the issue there is competing with plugins etc. that would use wp_footer (which is the action I use right now typically) and which sets the highest priority for an action hooked there.

Using a separate hook to place code at end of body tag seems cleaner to me.

As for use case, for example, some people prefer to put GA (or any other tracking) snippet at the very end in order to avoid counting visits from people that close browser window before page even finished loading.

lucspe:
Hi Leho,

Sorry to hear that you dont think it is 'much of an idea'. I thought it
would be useful and tried to explain why.

At any rate, thank you for sharing your point of view on this matter.


Il giorno ven 2 nov 2018, 19:20 Leho Kraav <notifications@github.com> ha
scritto:

lucspe:
Hi @tiagonoronha,

I had not thought of using PHP_INT_MAX as priority on wp_footer. That solved the need I had without the need for an additional action hook, indeed. 

Regards,
Luca

PR #1001, opened 2018-11-21
PHPCS errors and warnings

tiagonoronha:
This PR fixes all PHPCS errors and also most warnings. All the remaining warnings are, for the most part, related to the naming of a few class methods.

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -65,7 +65,7 @@ if ( post_password_required() ) {
 
 	endif;
 
-	if ( ! comments_open() && '0' != get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
+	if ( ! comments_open() && 0 !== get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
 		?>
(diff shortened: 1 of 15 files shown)


comments.php
```diff
@@ -65,7 +65,7 @@
 
 	endif;
 
-	if ( ! comments_open() && '0' != get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
+	if ( ! comments_open() && 0 !== get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
```
claudiosanches: `get_comments_number()` return `integer` only when is `0`, not a problem in this context, but just leaving this as a side note.

tiagonoronha:
@claudiosanches made the changes you requested.

As for `get_comments_number()`, it doesn't seem to be an issue in any of the places where we are using it. We are always either looking if it is actually 0 or something different than it :)

claudiosanches:
@tiagonoronha not an issue now, I'm just saying, that maybe we could leave a comment to not forget about that. But as I said, in my comment, just a side note. The problem is in WP returning mixed values there :(

PR #1002, opened 2018-11-22
Compatibility with Gutenberg editor

tiagonoronha:
This PR adds enhanced compatibility with the upcoming WordPress editor. Changes introduced in this PR:

* Declare support for Gutenberg features
* Styling for blocks included with the editor
* Change the single post layout from having the post meta to the left of the content to allow for blocks to go full width
* Add CSS file for future editor styles

**Post meta before:**

before

**Post meta after:**

after

### DIFF
--- a/inc/customizer/class-storefront-customizer.php
+++ b/inc/customizer/class-storefront-customizer.php
@@ -820,6 +820,78 @@ if ( ! class_exists( 'Storefront_Customizer' ) ) :
 		}
 
 		/**
+		 * Get Gutenberg Customizer css.
+		 *
+		 * @see get_storefront_theme_mods()
+		 * @return array $styles the css
+		 */
+		public function gutenberg_get_css() {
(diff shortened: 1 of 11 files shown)


inc/customizer/class-storefront-customizer.php
```diff
@@ -817,6 +819,82 @@ public function get_css() {
...
 
+		/**
+		 * Get Gutenberg Customizer css.
+		 *
+		 * @see get_storefront_theme_mods()
+		 * @return array $styles the css
+		 */
+		public function gutenberg_get_css() {
+			$storefront_theme_mods = $this->get_storefront_theme_mods();
+			$brighten_factor       = apply_filters( 'storefront_brighten_factor', 25 );
```
rodrigoprimo: Unused variable. I'm not sure if it should be removed or if it should be used in a similar fashion as `$darken_factor` is used.

PR #1003, opened 2018-11-23
Remove 'comments closed' message from pages

tiagonoronha:
Fixes the conditional logic for checking the number of comments. The issue was introduced in https://github.com/woocommerce/storefront/commit/a5f9c57962be2628b4f1d3c2ed5318612897fb58.

`get_comments_number()` returns a string, not an integer. @claudiosanches was right :)

Without this PR, you'll find the "comments closed" message at the bottom of pages. For example, check the Cart page and you'll find this:

screenshot 2018-11-23 at 11 44 11

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -65,7 +65,7 @@ if ( post_password_required() ) {
 
 	endif;
 
-	if ( ! comments_open() && '0' !== get_comments_number() && post_type_supports( get_post_type(), 'comments' ) ) :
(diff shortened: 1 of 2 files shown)


claudiosanches:
I personally prefer to force a type using `(int)` or `intval()`, so should avoid any problem with mexed values returned by WP.

tiagonoronha:
@claudiosanches decided to go with your suggestion and force the type with `intval()`.

Thank you!

rodrigoprimo (approved):
LGTM

PR #1012, opened 2018-12-06
Method incorrectly defined as static

### LINKED ISSUE #1011: [2.4] Backend does not load - WSOD
I'm running:
- WordPress 4.9.8
- Gutenberg 4.6.1
- Storefront 2.4

No plugins (except Gutenberg). I cannot enter the editor (gives the WSOD).

If I disable Gutenberg, the old editor loads fine. Other themes work fine. 

Also, if using WP 5 (and no Gutenberg plugin), the editor screen returns a 500 error, instead of the WSOD.

Known issues?

tiagonoronha:
The `block_editor_customizer_css()` method is currently incorrectly defined as `static`. This is causing fatal errors in some hosting platforms such as WPEngine.

Since there's no need for this method to be defined as `static`, this PR removes the keyword.

It also needs `global $storefront_version` which is not being used anywhere on the method.

Closes #1011.

### DIFF
--- a/inc/customizer/class-storefront-customizer.php
+++ b/inc/customizer/class-storefront-customizer.php
@@ -900,9 +900,7 @@ if ( ! class_exists( 'Storefront_Customizer' ) ) :
 		 *
 		 * @since 2.4.0
 		 */
-		public static function block_editor_customizer_css() {
-			global $storefront_version;
-
+		public function block_editor_customizer_css() {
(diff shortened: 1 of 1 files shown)


tiagonoronha:
@claudiulodro thanks, makes sense! opened a new issue for that: https://github.com/woocommerce/storefront/issues/1011

PR #1035, opened 2018-12-20
Refactor posted on sprintf

tiagonoronha:
Refactor posted meta `sprintf` output and separate the different elements for a better readability.

The output should stay the same in posts:

screenshot 2018-12-20 at 16 37 51


tiagonoronha:
@claudiulodro your suggestion makes sense, update the PR. :)

PR #1037, opened 2018-12-29
Update storefront-template-functions.php

tazotodua:
We switched our theme to storefront, but I was a bit surprised that this theme didn't support customization more than it is at this moment, for example, we want to override the output from "default" hardcoded functions. there seems to be no filters at all. The only ability was to redefine the functions on the run-time (which is quite inflexible, for example we want to control the filter during execution to conditionally control the output, which is impossible with functions.  Also, it's impossible with functions, that they were overriden from 2 different plugins, because the "function already defined" problems arise.  So, it's better to use wordpress filters).
It would be nice if in addition of this modification, you later implemented filters to control the output of `echo` commands too.

### DIFF
--- a/inc/storefront-template-functions.php
+++ b/inc/storefront-template-functions.php
@@ -12,6 +12,9 @@ if ( ! function_exists( 'storefront_display_comments' ) ) {
 	 * @since  1.0.0
 	 */
 	function storefront_display_comments() {
+		if ( ! apply_filters('storefront_display_comments', true) ) 
+			return;
+		
 		// If comments are open or we have at least one comment, load up the comment template.
 		if ( comments_open() || 0 !== intval( get_comments_number() ) ) :
 			comments_template();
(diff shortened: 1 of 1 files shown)


tiagonoronha:
@tazotodua thank you for your contribution.

I'm not sure I understand why this is needed. Functions in the `storefront-template-functions.php` file are functions we use in hooks. You can override these functions by:

1) Unhooking the function with `remove_action()` and then replacing it with something else (or not). Have a look at the `storefront-template-hooks.php` and `storefront-woocommerce-template-hooks.php` file so you can see where these functions are hooked.
2) Redeclaring the function in a child theme. You won't get any warnings because of the conditional check using `function_exists()`.

I'm closing this PR.

tazotodua:
I am worried you denied the commit. Well, I'd suggest that you consulted with other pros or you parnters too, so, you might easier understand how big difference is between `if(function_exists)` override, or overriding with `apply_filters`. also, `apply_filters` doesnt have any negative side. 
So, what you suggest is that for the simple modification, we have to either create a child theme, or  copy paste all functions, in order to make a modification into them. this is what i was surprised. 

the best thing was that we could use `apply_filters`, and I don't know why you don't use apply_filters, just i dont get the point, if you used them, there was no negative effect, and we could control them from our PLUGINS too (not from themes).

PR #1047, opened 2019-01-25
Remove padding and transition styles from post navigation

tiagonoronha:
Removes padding and hover transition from the post navigation so it sits flush at the edge of the main content area.

To test, visit a regular post and look for the post navigation links at the end of the post.

Before:

screenshot 2019-01-25 at 19 12 40

After:

screenshot 2019-01-25 at 19 13 39

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -1296,33 +1296,18 @@ button.menu-toggle {
 	.nav-next {
 		a {
 			display: inline-block;
-			transition: all, ease, 0.3s;
-			padding: ms(-2) 1em;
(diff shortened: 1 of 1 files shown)


GlennMartin1:
I assume this change would not conflict with the Homestore child-theme that shows a small arrow that extends into the padding area that you're removing?

Perhaps it's fine - just noting it....

tiagonoronha:
Yup, I can confirm that this change does make the Homestore navigation links look a bit funky.

I'm going to merge this PR, but then update Homestore and other child themes that might require styling adjustments.

PR #1060, opened 2019-02-13
Updated node dependencies

claudiosanches:
Updating all our node dependencies, also this applies new security patches for several of the dependencies.


tiagonoronha:
@claudiosanches try to run `npm install` and `npm run build`. Do you get errors with this PR?

claudiosanches:
@tiagonoronha I had to revert susy version, since there is a few removed functions without any guide to migrate right now, probably better let to update this later.

tiagonoronha:
I replaced `node-bourbon` with the official version.

There's a bunch of deprecation warnings now coming from Susy, so we'll have to update that eventually.

PR #1068, opened 2019-03-02
Minify base css

jaylinski:
I'm currently doing some performance optimizations on my shop and noticed the unminified icon-styles.

Is there a specific reason why the CSS in the `base` folder is not minified?


tiagonoronha:
Thank you!

PR #1069, opened 2019-03-02
Minify vendor JS

jaylinski:
The existing minification of the `pep.min.js` file was poor.   
By using our own uglify script, we can reduce the size by 6KB.

Also replaced a deprecated uglify option.   
The `preserveComments` is deprecated since grunt-contrib-uglify 2.x.


assets/js/vendor/pep.js
```diff
@@ -207,4 +207,4 @@ dedupSynthMouse:function(a){var b=N.lastTouches,c=a.changedTouches[0];
...
...cloneEvent(a),b.pointerType=this.POINTER_TYPES[a.pointerType]),b},cleanup:function(a){Z["delete"](a)},MSPointerDown:function(a){Z.set(a.pointerId,a);var b=this.prepareEvent(a);u.down(b)},MSPointerMove:function(a){var b=this.prepareEvent(a);u.move(b)},MSPointerUp:function(a){var b=this.prepareEvent(a);u.up(b),this.cleanup(a.pointerId)},MSPointerOut:function(a){var b=this.prepareEvent(a);u.leaveOut(b)},MSPointerOver:function(a){var b=this.prepareEvent(a);u.enterOver(b)},MSPointerCancel:function(a){var b=this.prepareEvent(a);u.cancel(b),this.cleanup(a.pointerId)},MSLostPointerCapture:function(a){var b=u.makeEvent("lostpointercapture",a);u.dispatchEvent(b)},MSGotPointerCapture:function(a){var b=u.makeEvent("gotpointercapture",a);u.dispatchEvent(b)}},aa=window.navigator;aa.msPointerEnabled?(W=function(a){i(a),j(this),k(a)&&(u.setCapture(a,this,!0),this.msSetPointerCapture(a))},X=function(a){i(a),u.releaseCapture(a,!0),this.msReleasePointerCapture(a)}):(W=function(a){i(a),j(this),k(a)&&u.setCapture(a,this)},X=function(a){i(a),u.releaseCapture(a)}),Y=function(a){return!!u.captureInfo[a]},g(),h(),l();var ba={dispatcher:u,Installer:c,PointerEvent:a,PointerMap:p,targetFinding:v};return ba});
```
jaylinski: I can remove this "no new line at end of file" change. But it's due to the `insert_final_newline` setting inside the editorconfig.

assets/js/vendor/pep.js
```diff
@@ -1,6 +1,6 @@
 /*!
  * PEP v0.4.3 | https://github.com/jquery/PEP
- * Copyright jQuery Foundation and other contributors | http://jquery.org/license
+ * @license Copyright jQuery Foundation and other contributors | http://jquery.org/license
```
jaylinski: The `@license` keyword will preserve the comment when the file is uglified.

tiagonoronha:
👍 Thank you!

jaylinski:
@tiagonoronha Because if a file contains `min`, it won't be uglified:
https://github.com/woocommerce/storefront/blob/dd6a19c81b866149a8db915fe14f64b3c4cae169/Gruntfile.js#L87

I know that renaming the file from `pep.min.js` to `pep.js` is not ideal, because the file is already somewhat minified, but it was the easiest way to get it uglified by the grunt-script...

tiagonoronha:
This is a static file, there's no need for it to be minified every time we run grunt. Please minify it and then change the name back to `pep.min.js`. Thank you!

jaylinski:
@tiagonoronha You are right! Done. :+1:

tiagonoronha:
Thank you :)

PR #1074, opened 2019-03-07
Fix handheld navigation accessibility

### LINKED ISSUE #1070: Assistive tech (VoiceOver) not reading hamburger menu correctly
As reported in #1836944-zen the expectation with VoiceOver and  collapsed hamburger menu is that tabbing through screen elements would skip the menu items, however, this is not the case i.e. menu items that exist are still read by the screen reader.

The expected behavior can be seen with the Twenty Seventeen theme.

⌘ + F5 will turn on VoiceOver. Tab around the site to listen to the descriptions

You can use Google Chrome inspect to visit the mobile view, and see the hamburger menu. Tab through to hear it speak each menu item, even when the menu remains collapsed.

This can be fixed by toggling div.handheld-navigation display on the same event that toggles the button https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element

tiagonoronha:
When using VoiceOver, the handheld menu was still being read even though it wasn't visible on the page. This was because we set the `min-height` on the menu to `0` on load.

Without this PR, VoiceOver will still read the menu because technically it is there and not hidden.

* Hides handheld menu with CSS `visibility`
* Add `area-expanded` attribute to correct sections
* Add `aria-haspopup` attribute to indicated a sub menu is present

I moved the order of some of the JS inside `navigation.js`. The diff for this PR looks bigger than the actual changes.

How to test:

* Go to WordPress > Menus and add a few items to the "Handheld" location.
* Resize your browser until the mobile nav kicks in and turn on VoiceOver (⌘ + F5)
* Tab around the site to listen to the descriptions
* After you get to the "Menu" button, it should skip to the next section and not read the hidden collapsed menu.

Closes #1070.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -1256,7 +1256,7 @@ button.menu-toggle {
 .main-navigation.toggled {
 	.handheld-navigation,
 	.menu > ul:not( .nav-menu ),
-	ul[aria-expanded=true] {
+	.handheld-navigation > ul.nav-menu {
 		max-height: 9999px;
 	}
 }
--- a/assets/js/navigation.js
+++ b/assets/js/navigation.js
@@ -11,6 +11,7 @@
 	// Wait for DOM to be ready.
 	document.addEventListener( 'DOMContentLoaded', function() {
 		var container = document.getElementById( 'site-navigation' );
+
 		if ( ! container ) {
 			return;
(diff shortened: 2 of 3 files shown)


haszari:
This PR was reverted: #1081

PR #1162, opened 2019-06-11
Fix uneven form input widths on checkout page

### LINKED ISSUE #1144: Form inputs with uneven widths
`.col2-set .form-row-first` and `.col2-set .form-row-last` have very different widths in the CSS yielding uneven measures.

```css
.col2-set .form-row-first {
	width: 33.3333333333%;
	float: left;
	margin-right: 11.1111111111%;
}
```
```css
.col2-set .form-row-last {
	width: 55.5555555556%;
	float: right;
	margin-right: 0;
}
```

image

malviyaritesh:
On checkout page first name and last name form inputs were of
uneven widths. Fix #1144 by making them of equal width.

### DIFF
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -2564,11 +2564,11 @@ dl.variation {
 		@include span(5 of 9);
 
 		.form-row-first {
-			@include span(2 of 5);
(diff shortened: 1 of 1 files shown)


tiagonoronha:
Looks good! :) 

Thank you!

PR #1163, opened 2019-06-16
Added input type styles to the 'select' type

robertdevore:
I was recently working on a plugin that adds a select field to the Checkout form and noticed that the stylesheet doesn't have the select type added to it's CSS.

Before:

storefront-select-before

After:

storefront-select-after

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -935,6 +935,7 @@ input[type='tel'],
 input[type='url'],
 input[type='password'],
 input[type='search'],
+input[type='select'],
 textarea,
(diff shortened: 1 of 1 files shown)


tiagonoronha:
Hi @robertdevore! 

Thank you for your contribution.

In your PR you're targeting `input[type='select']` which is not a valid HTML element. I'm guessing that you meant simply `select`.

For accessibility reasons we are keeping the selects as they are for now. For example, in your image above, the select is missing the dropdown arrow which makes it nearly impossible to understand that the field is a select box.

PR #1166, opened 2019-06-17
Fix cover block paragraph size and default color

### LINKED ISSUE #1153: Cover block - incorrect default text color
The default text colour rendered on the block is different from what is rendered in the preview.

Screenshot 2019-05-30 at 06 38 25

tiagonoronha:
This PR fixes two issues related to the cover block:

* Font size was fixed at `1.1em`
* Default paragraph color not matching frontend color

### To test:

 **Font size was fixed to `1.1em`**

* Add a cover block
* Add a paragraph inside the cover and add some text
* Default font size should be the same in the editor and frontend
* In the editor sidebar change the size of the paragraph text (Small, Medium, Large, Huge). Size should be the same in the editor and frontend.

**Default paragraph color not matching frontend color**

* Add a cover block
* Add a paragraph inside the cover and add some text
* By default the text color for paragraphs should be **white** both in the editor and then in the frontend
* In the editor sidebar change the color of the paragraph. Ensure the color changes in the editor and frontend

Closes #1153.

### DIFF
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -406,11 +406,12 @@
 			}
 
 			p {
-				&:not( .has-small-font-size ),
-				&:not( .has-medium-font-size ),
-				&:not( .has-large-font-size ),
-				&:not( .has-huge-font-size ) {
+				&:not( .has-small-font-size ):not( .has-medium-font-size ):not( .has-large-font-size ):not( .has-huge-font-size ) {
 					font-size: 1.1em;
+				}
+
+				&:not( .has-text-color ) {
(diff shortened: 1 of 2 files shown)


tiagonoronha:
@rodrigoprimo good catch!

I've now added Storefront's custom sizes to the editor via `add_theme_support()` in https://github.com/woocommerce/storefront/pull/1166/commits/a2aec78b8b200e5fe65a70fb61d7b91cc86d3a25.

It's almost 1:1, but since the editor only supports `px`, and Storefront uses `em` for sizes, there is a ~.5/1px difference in the sizes. But it's not as noticeable as before.

PR #1178, opened 2019-07-04
Move privacy link outside credit link conditional

### LINKED ISSUE #1175: store_credit_link conditional incorrectly wraps the privacy link as well as the credit link. (version 2.5.0)
In version 2.5.0 of the storefront theme, storefront/inc/storefront-template-functions.php

function storefront_credit()

This conditional:   <?php if ( apply_filters( 'storefront_credit_link', true ) ) { ?>

is incorrectly wrapping the_privacy_policy_link as well as the storefront_credit_link.

It should be possible to disable the credit link without also disabling the privacy link, but as the code is written, it isn't possible without adding a replacement storefront_credit() function in the child theme's functions.php.

Thanks,
John Garnett

tiagonoronha:
Right now it is not possible to remove the Storefront credits link in the footer without also removing the privacy link.

This PR separates separates both pieces so that they can be toggled on/off independently. It also adds a bit of extra logic in case there's no credit link, in that case we don't want the separator ` |  ` to show up.

To test, try to toggle the filters on/off and see if everything displays as expected:

```
add_filter( 'storefront_credit_link', '__return_false' );
```

```
add_filter( 'storefront_privacy_policy_link', '__return_false' );
```

Closes #1175.

### DIFF
--- a/inc/storefront-template-functions.php
+++ b/inc/storefront-template-functions.php
@@ -134,17 +134,23 @@ if ( ! function_exists( 'storefront_credit' ) ) {
 	 * @return void
 	 */
 	function storefront_credit() {
+		$links_output = '';
+
+		if ( apply_filters( 'storefront_credit_link', true ) ) {
+			$links_output .= '<a href="https://woocommerce.com" target="_blank" title="' . esc_attr__( 'WooCommerce - The Best eCommerce Platform for WordPress', 'storefront' ) . '" rel="author">' . esc_html__( 'Built with Storefront &amp; WooCommerce', 'storefront' ) . '</a>.';
+		}
+
+		if ( apply_filters( 'storefront_privacy_policy_link', true ) && function_exists( 'the_privacy_policy_link' ) ) {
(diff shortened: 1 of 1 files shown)


sisou:
With this change, the call to the `storefront_privacy_policy_link` filter is not inside the `.site-info` span anymore, so this breaks _adding_ a link to the footer in this section with this filter.

I was using it to add a link to the Terms & Conditions in front of the privacy policy in this location by simply `echo`ing the link code in the `storefront_privacy_policy_link` filter. This now puts the HTML code _outside_ the `.site-info`, breaking the footer.

sisou:
My proposal would be to keep the current separation between the filters, but move the PHP code _inside_ the `.site-info` div, so it's executed at the correct HTML DOM position.

peterfabian:
Hi @sisou, thanks for your comment. 
Could you please create a new issue in this repo and link this PR so that we don't foget about this and can look into it as a new task?

PR #1183, opened 2019-07-15
Move calls to storefront_credit filters to their rendering location

### LINKED ISSUE #1128: [2.5] console errors...
Wordpress 5.2.0
Storefront 2.5
Homestore 2.0.31

On all my installation with this environment, I get this console error:

console errors

### LINKED ISSUE #1182: Adding content with `storefront_privacy_policy_link` filter regression
With #1178, the call to the `storefront_privacy_policy_link` filter is not inside the `.site-info` div anymore, so this breaks _adding_ a link to the footer in this section with this filter.

I was using it to add a link to the Terms & Conditions in front of the privacy policy in this location by simply `echo`ing the link code in the `storefront_privacy_policy_link` filter. This now puts the HTML code _outside_ the `.site-info`, breaking the footer.

My proposal would be to keep the current separation between the filters, but move the PHP code _inside_ the `.site-info` div, so it's executed at the correct HTML DOM position.

sisou:
Fixes #1182

This moves the calls to the `storefront_credit_link` and `storefront_privacy_policy_link` filters back inside the `.site-info` div, so those filters are called at the location where they are rendered.

This keeps the separation of the filters introduced in #1178 while also preserving the usage of _adding_ content to this location by `echo`ing it from one of the filters.

I had to replace the `<br />` linebreak with a `<div>`, because the linebreak would have always be rendered after the filter call. The DIV has the same functionality here (breaking into a new line) and does not take up space when both filters are disabled (it just becomes a zero-height DIV then).

### DIFF
--- a/inc/storefront-template-functions.php
+++ b/inc/storefront-template-functions.php
@@ -134,24 +134,26 @@ if ( ! function_exists( 'storefront_credit' ) ) {
 	 * @return void
 	 */
 	function storefront_credit() {
-		$links_output = '';
-
-		if ( apply_filters( 'storefront_credit_link', true ) ) {
-			$links_output .= '<a href="https://woocommerce.com" target="_blank" title="' . esc_attr__( 'WooCommerce - The Best eCommerce Platform for WordPress', 'storefront' ) . '" rel="author">' . esc_html__( 'Built with Storefront &amp; WooCommerce', 'storefront' ) . '</a>.';
-		}
-
-		if ( apply_filters( 'storefront_privacy_policy_link', true ) && function_exists( 'the_privacy_policy_link' ) ) {
-			$separator = '<span role="separator" aria-hidden="true"></span>';
-			$links_output = get_the_privacy_policy_link( '', ( ! empty( $links_output ) ? $separator : '' ) ) . $links_output;
-		}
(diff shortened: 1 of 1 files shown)


tiagonoronha:
Hi @sisou. Outputting inside a filter is not a very good practice and can lead to very unexpected results.

The ideal solution would be to add a new filter to `$links_output`.

`$links_output = apply_filters( 'storefront_credit_links_output', $links_output );`

You could then filter the output to add your custom link.

sisou:
I am also OK with adding a new filter, however I have no idea how that would work. Is there more required than simply calling the `apply_filters` method?

If you think the additional filter is the way to go, feel free to close this PR ;)

vedanshujain:
@sisou yup its mostly `apply_filter` method. So basically just above where we do `echo wp_kses_post( $links_output );`, you would add one more line like so `$links_output = apply_filters( 'storefront_credit_links_output', $links_output );`, and then hook into this newly added filter to add the content that you want to.
I am closing this PR, I would request you to submit another PR for the new filter, or create a new issue if you won't have time to submit a PR.

sisou:
I created a PR that adds the filter at #1191.

PR #1186, opened 2019-07-23
Smoothout the footer layout

olSkoolMarco:
Just wanted to make to footer more attractive.
Author and link remains the same, changed style from the text to image.
Also added style to image, so it can be re-adjusted if required.


haszari:
This PR includes an arbitrary change in appearance to the footer, to use an image instead of text for the WooCommerce & Storefront credit. This is not an essential bug fix or improvement, so closing this PR.

PR #1187, opened 2019-07-26
Don't apply link styles to core components-button

simison:
New WordPress components (in [`@wordpress/components`](https://github.com/WordPress/gutenberg/tree/0779690cf4345a2468e294cbab0c8f6b805073fb/packages/components)) have a button with CSS selector [`.components-button`](https://github.com/WordPress/gutenberg/blob/0779690cf4345a2468e294cbab0c8f6b805073fb/packages/components/src/button/style.scss#L1).

This PR adds `:not` exception for all link stylings in addition to existing `:not( .button )` selector.

I'm adding multiple `:not()` selectors because of support for multiple selectors inside `:not()` is [still lacking](https://caniuse.com/#feat=css-not-sel-list).

### Testing

Insert following HTML to a post:
```html
<a href="#" class="components-button is-button is-default"><span>Test</span></a>
```

And enqueue style from Gutenberg in PHP: `wp_enqueue_style( 'wp-components' );`

Confirm there is no underline in the test button:
image
image

Came up via https://github.com/Automattic/jetpack/pull/13070#discussion_r307714522

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -455,7 +455,7 @@ body {
 		color: mix( $color_body, #222 );
 	}
 
-	a:not( .button ):not( .component-button ) {
+	a:not( .button ):not( .components-button ) {
 		color: $color_links;
 		text-decoration: underline;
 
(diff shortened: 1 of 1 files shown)


haszari:
Added this to 2.5.4 as it's a low risk tightly-scoped change; would be good to get this merged soon.

PR #1189, opened 2019-08-02
Introduce extra block styles

mikejolley:
Adds extra style rules targeting the new markup within blocks. This also fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/795.

Requires https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/796 which reduces CSS specificity to allow these rules to take precedence. 

Before:
Screenshot 2019-08-02 at 12 47 36

After:
Screenshot 2019-08-02 at 12 48 12

Not sure who is managing SF to review @peterfabian.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -848,13 +848,20 @@ textarea {
 	/* stylelint-enable */
 }
 
+/* Override gutenberg styles. */
+.wp-block-button__link {
+	border-radius: 0 !important;
+}
+
 button,
 input[type='button'],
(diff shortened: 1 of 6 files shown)


mikejolley:
@rodrigoprimo I've just pushed a fix for those 2 styles. **However**, this should have a more in-depth look at some point. We're needing to duplicate frontend styles in the editor stylesheet, and it's likely this will cause conflicts in the future. I'm not sure of the best approach here, but possible a shared stylesheet will be needed.

PR #1199, opened 2019-08-20
Exclude storefront/ and storefront.zip when preparing a release

rodrigoprimo:
This PR makes sure that storefront/ and storefront.zip are not included in the list of files that are copied to the directory used to create a production ready version of the theme using the command `npm run deploy`. This is necessary to avoid issues like the one that happened in the 2.5.2 version where zip file generated for the release included a duplicated copy of Storefront inside it.

I also included a separate commit to sort the list of files to copy alphabetically to make it easier to make changes in the future.

**How to test**
- On the master branch, run the command `npm run deploy` twice and see that the generated `storefront` directory contains another `storefront` directory inside it (`repo_root/storefront/storefront`) and also a copy of `storefront.zip`.
- Switch to this branch, remove the `storefront` directory and run the same command twice again. Check that there isn't a second `storefront` directory or a `storefront.zip` file inside `repo_root/storefront/`.
- Compare the contents of the `storefront/` directory generated before and after this change to make sure they are equal. The only difference should be the generated time in the .pot file. I used a local git repository to compare both directories.


haszari:
Added this to 2.5.4 – I think it will make the release process more robust.

Issue #1197, opened 2019-08-20
WC Products by Category block does not respect columns


ajgw:
If I create a products by category block into post/page, last product in first row breaks into next row. Following products are as expected. This is related to Storefront 2.5.3.

![Bildschirmfoto 2019-08-20 um 13 23 29](https://user-images.githubusercontent.com/3390880/63343212-b77f9280-c34d-11e9-87d0-c3463f02f928.png)

vedanshujain:
I am not able to reproduce on a fresh StoreFront + WooCommerce install. Can you please provide a System report from Dashboard -> WooCommerce -> Status. 
Can you also specify what is your screen resolution, this can help us in reproducing the issue.

ajgw:
Well, I gave it another test on a sting site. I deactivated all plugins except Woocommerce. I have setup only one block Products by Category with setting 3 columns, 2 rows. I had Woocommerce 3.6.5 and Storefront 2.5.1 on site and did a screenshot at start of test, after update Woocommerce to 3.7.0 and last after update of Storefront to 2.5.3:

![Woocommerce 3-6-5 and Storefront 2-5-1](https://user-images.githubusercontent.com/3390880/63538730-c5cdda00-c518-11e9-8bc7-216c742db322.png)

![Woocommerce 3-7-0 and Storefront 2-5-1](https://user-images.githubusercontent.com/3390880/63538750-d5e5b980-c518-11e9-9994-25f3aa568151.png)

![Woocommerce 3-7-0 and Storefront 2-5-3](https://user-images.githubusercontent.com/3390880/63538769-dd0cc780-c518-11e9-8a06-a18170348cc3.png)

ajgw:
You can see in first row that last product is not in row, it is broken into next row.

ajgw:
Screenresolution is Macbook Pro 15 inch Retina = 2880 by 1800 px. Systemreport:

`
### WordPress Environment ###

WordPress address (URL): https://stage.dekuenst.de
Site address (URL): https://stage.dekuenst.de
WC Version: 3.7.0
REST API Version: ✔ 1.0.2
Log Directory Writable: ✔
WP Version: 5.2.2
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: –
WP Cron: ✔
Language: de_DE
External object cache: –

### Server Environment ###

Server Info: Apache
PHP Version: 7.3.6-nmm1
PHP Post Max Size: 200 MB
PHP Time Limit: 30
[...]

PR #1200, opened 2019-08-26
Override content:'' from included clearfix to none.

vedanshujain:
We were including `content: ''` from `@include clearfix;` which was causing issues in Safari. Adding a rule to override the specific rule which was causing issues.

Fixes https://github.com/woocommerce/storefront/issues/1197.


vedanshujain:
Closing in favor of #1217

PR #1206, opened 2019-09-10
Add link text on cart button in handheld footer bar (fixes #1205)

### LINKED ISSUE #1205: Missing link text on cart button in handheld footer bar
All buttons in handheld footer bar have a link text, except the "cart" button. Also it has a title attribute which is not used on mobile touch devices.

https://github.com/woocommerce/storefront/blob/03b9229f0d5f3924ffbfa13d0fd7fe2a085e6338/inc/woocommerce/storefront-woocommerce-template-functions.php#L682

. 
Having a link text on all buttons allows web designer to remove text indent and show text below icons

footerbar

uok:
Fixes #1205

- add link text on cart button
- remove title attribute

### DIFF
--- a/inc/woocommerce/storefront-woocommerce-template-functions.php
+++ b/inc/woocommerce/storefront-woocommerce-template-functions.php
@@ -679,7 +679,7 @@ if ( ! function_exists( 'storefront_handheld_footer_bar_cart_link' ) ) {
 	 */
 	function storefront_handheld_footer_bar_cart_link() {
 		?>
-			<a class="footer-cart-contents" href="<?php echo esc_url( wc_get_cart_url() ); ?>" title="<?php esc_attr_e( 'View your shopping cart', 'storefront' ); ?>">
+			<a class="footer-cart-contents" href="<?php echo esc_url( wc_get_cart_url() ); ?>"><?php esc_attr_e( 'Cart', 'woocommerce' ); ?>
 				<span class="count"><?php echo wp_kses_data( WC()->cart->get_cart_contents_count() ); ?></span>
 			</a>
 		<?php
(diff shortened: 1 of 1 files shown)


haszari:
Added `Accessibility` label, as this probably [improves accessibility of the footer cart link](https://www.w3.org/TR/WCAG20-TECHS/H30.html) 🚀 .

inc/woocommerce/storefront-woocommerce-template-functions.php
```diff
@@ -679,7 +679,7 @@ function storefront_handheld_footer_bar_search() {
 	 */
 	function storefront_handheld_footer_bar_cart_link() {
 		?>
-			<a class="footer-cart-contents" href="<?php echo esc_url( wc_get_cart_url() ); ?>" title="<?php esc_attr_e( 'View your shopping cart', 'storefront' ); ?>">
+			<a class="footer-cart-contents" href="<?php echo esc_url( wc_get_cart_url() ); ?>"><?php esc_attr_e( 'Cart', 'woocommerce' ); ?>
```
Aljullu: We need to keep the same translation domain name and the escaping function should be now `esc_html_e()` instead of `esc_attr_e()` because it's no longer an attribute.
```suggestion
			<a class="footer-cart-contents" href="<?php echo esc_url( wc_get_cart_url() ); ?>"><?php esc_html_e( 'Cart', 'storefront' ); ?>
```

Aljullu (commented):
Thanks for opening this PR @uok!

I tested with a screen reader and the text that is read changes from `View your shopping cart` to `Cart`. I assume that's ok considering it's a link.

Visually, I couldn't spot any difference in Storefront neither in a couple of child themes I tested (Arcade and Bistro), which is great. :ok_hand: 

I only left a coding comment below which would be great to address before merging. :point_down:

EdithAllison:
This change made our child theme look odd. If anyone wants to hide the "cart" text but wishes to continue using the storefront function (which ties into the fragment ajax refresh), then a simple CSS solution is to

- set the parent a.footer-cart-contents to "visibility: hidden;"
-  set the children a.footer-cart-contents::before and span.count to "visibility: visible;"

This will hide the "Cart" text from view & make the link appear as before.

PR #1207, opened 2019-09-23
Prevent indexing of add to cart link

ntindicator:
Search engines are indexing the add to cart link from this tag. This doesn't occur within Woocommerce as an <a> tag isn't used for the button.


haszari:
Added this to 2.5.4 since it's low risk and approved :)

Thanks for the help @ntindicator!

Issue #820, opened 2018-04-12
Use Core Screen Reader Classes


tech55541:
Some of the screen-reader-text class styling is very messy. Making some words read with no spacing. "Previouspost:example". Please update the screen-reader-text class to conform with the core class.
https://make.wordpress.org/core/2017/10/22/changes-to-the-screen-reader-text-css-class-in-wordpress-4-9/

I could not find a way to build the sass myself for a PR.

jobthomas:
Just adding this from the forum thread: https://wordpress.org/support/topic/add-screen-reader-label-to-post-navigation/

> Hello,
> 
> Could you please add a screen reader label to post navigation links?
> 
> ```
> if ( ! function_exists( 'storefront_post_nav' ) ) {
> 	/**
> 	 * Display navigation to next/previous post when applicable.
> 	 */
> 	function storefront_post_nav() {
> 		$args = array(
> 			'next_text' => '<span class="screen-reader-text">Next post: </span>%title',
> 			'prev_text' => '<span class="screen-reader-text">Previous post: </span>%title',
> 			);
> 		the_post_navigation( $args );
> 	}
> }
> ```
> 
> Thanks.

tiagonoronha:
Hi @tech55541! Thanks for reporting this issue. :)

In case you're still interested in submitting a PR, Instructions on how to build the Sass files are available here: https://github.com/woocommerce/storefront/blob/master/CONTRIBUTING.md

tech55541:
Pull request created.

@tiagonoronha

WraithKenny:
This merge broke the ability to focus `.screen-reader-text.skip-link` visually.

PR #1216, opened 2019-10-08
Make skip-link focusable, keyboard accessibility

WraithKenny:
This is a simple patch. Removing `clip-path` on focus fixes the issue.

We don't want ***all*** `.screen-reader-text` to be visible on focus, but we do want `.screen-reader-text.skip-link` to be visible.


WraithKenny:
See #820 and #824 for history. This is just a simple amend to that, to fix the focus issue.

PR #1241, opened 2019-12-27
#1208: Add support for the wp_body_open hook

### LINKED ISSUE #1208: Add support for the wp_body_open hook
A recommended hook added in WordPress 5.2 - https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/

nielslange:
Fixes #1208 

Change: Adds the `wp_body_open` hook after the opening body tag and added backwards compatibility to it via a separate file. This change is the result of the introduced `wp_body_open` hook in https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/.

### DIFF
--- a/inc/wordpress-shims.php
+++ b/inc/wordpress-shims.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * WordPress shims.
+ *
+ * @package storefront
+ */
+
+if ( ! function_exists( 'wp_body_open' ) ) {
+    /**
+     * Adds backwards compatibility for wp_body_open() introduced with WordPress 5.2
+     *
+     * @since 2.5.3
+     * @see https://developer.wordpress.org/reference/functions/wp_body_open/
+     * @return void
+     */
+    function wp_body_open() {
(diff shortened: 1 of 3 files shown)


inc/wordpress-shims.php
```diff
@@ -0,0 +1,19 @@
...
+ * WordPress shims.
+ *
+ * @package storefront
+ */
+
+if ( ! function_exists( 'wp_body_open' ) ) {
+    /**
+     * Adds backwards compatibility for wp_body_open() introduced with WordPress 5.2
+     *
+     * @since 2.5.3
```
Aljullu: A release was done after this PR was created, so we will need to update this line to `2.5.4`.
```suggestion
     * @since 2.5.4
```

haszari: @nielslange can you apply this change on your branch? Then we can merge this :)

nielslange: @haszari The version number had been adjusted according to your feedback. 😀

PR #1242, opened 2019-12-27
#1230: Image block options left aligned in Storefront

### LINKED ISSUE #1230: Image block options left aligned in Storefront
Image block options are left aligned when the Storefront theme is activated.

### Steps to reproduce

1. I added an Image block to a post on my self-hosted WordPress.org site, https://pljcbsntest1.mystagingwebsite.com, with Storefront active.
2. I noticed that the options for the Image block are left aligned, [like this](https://d.pr/i/VQckP3/oXId5t3djT).
3. Switching to the Twenty Twenty theme results in the Image block options being center aligned, as I'd expect, [like this](https://d.pr/i/IakE8v/o4bqv4oj3M).

### What I expected

Image block options ought to be center aligned, regardless of the active theme, [like this](https://d.pr/i/IakE8v/o4bqv4oj3M):

_Screenshot direct link:_ https://d.pr/i/IakE8v/o4bqv4oj3M

Instead, they're left aligned with Storefront active, [like this](https://d.pr/i/VQckP3/oXId5t3djT):

_Screenshot direct link:_ https://d.pr/i/VQckP3/oXId5t3djT

### Observations

I reactivated Storefront, and checked the CSS that applies to the options, and noticed that when I disable `max-width: 400px;` for the `.components-placeholder__fieldset, .components-placeholder__fieldset form` element, the block options display correctly, [like this](https://d.pr/i/adF5rP/3DC2SOjND9):

_Screenshot direct link:_ https://d.pr/i/adF5rP/3DC2SOjND9

@garymurray also did some testing, and noted the following:

>I tested it and can confirm the issue – however when you have the Twenty Twenty theme activated I can see the same `max-width: 400px;` applied to the `.components-placeholder__fieldset, .components-placeholder__fieldset form` element. 

### Browser / OS version

I was using Firefox Developer Edition, v71.0b12 on macOS 10.15.1

#### Context / Source

Internal testing of the Jetpack Beta v8.0.0 plugin.

nielslange:
Fixes #1230

<table>
<tr>
<td>Before:
<br><br>

#1230-before

</td>
<td>After:
<br><br>

#1230-after

</td>
</tr>
</table>

### DIFF
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -330,10 +330,6 @@
 		margin: 0 0 ms(2);
 	}
 
-	div.wp-block-image {
-		display: inline;
-	}
-
 	// Cover
 	.wp-block-cover {
 		// < 5.2 styling


assets/css/base/gutenberg-blocks.scss
```diff
@@ -330,10 +330,6 @@
 		margin: 0 0 ms(2);
 	}
 
-	div.wp-block-image {
-		display: inline;
-	}
-
```
haszari: This was added to "fix floated image alignment", so will need to test to ensure floated images still work correctly.

nielslange: Shouldn't the CSS code for floated images be adjusted separately? It obviously broke the layout of the image block as seen on the screenshot above.

haszari: Yes, we'll need to find a way to target these css rules appropriately. I tested all the align options for the image block with Storefront to see if there are any regressions. With this PR, the text flows (wraps) around `right`/`left` aligned image blocks in editor, but not in front end. 

Screen Shot 2020-01-31 at 10 54 30 AM

In the current released version of Storefront, text does not wrap in front end or editor, so we'll need to resolve that. I'll take a closer look at this and see if we need to target the editor buttons specifically, and/or if we have a bug in left/right align that we need to log & fix.

haszari: Tested the align options and this fix is working well, I couldn't find any regressions.

PR #1246, opened 2020-01-07
Update header.php to replace storefront_before_site with wp_body_open

lucspe:
WordPress 5.2 introduces natively the 'wp_body_open' action hook, that essentially serves the same purpose of 'storefront_before_site' introduced in 2.2.4.
It might be worthwhile (for greater compatibility with plugins that can use that hook) to consider replacing 'storefront_before_site' hook with the standardized one.

See: https://make.wordpress.org/core/2019/04/24/miscellaneous-developer-updates-in-5-2/

### DIFF
--- a/header.php
+++ b/header.php
@@ -20,7 +20,14 @@
 
 <body <?php body_class(); ?>>
 
-<?php do_action( 'storefront_before_site' ); ?>
+<?php 
+	<?php 
+	if ( function_exists( 'wp_body_open' ) ) {
+		wp_body_open();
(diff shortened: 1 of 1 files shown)


haszari:
Thanks for your input @lucspe :)

We have a PR adding support for `wp_body_open` here, using the standard approach: https://github.com/woocommerce/storefront/pull/1241

We should keep `storefront_before_site` hook for backward compatibility. 

With that in mind, I'm closing this PR.

PR #1251, opened 2020-01-28
All Products block: don't make sale badge transparent to avoid constrast issues

Aljullu:
### Screenshots

imatge

1. Make sure you have at least one product on sale.
2. Create a post and add the _All Products_ block.
3. Observe it in the frontend and verify the `Sale` badge has white background. Before, it was transparent and it would cause accessibility issues with dark images.


Aljullu:
Good catch @haszari! Fixed in the last commit. :+1:

Issue #1181, opened 2019-07-15
If text color and heading color are set to white in Appearance >> Customize >> Typography, the text is also white when editing pages making the text invisible


grigorijurasov:
**Describe the bug**
If text color and heading color are set to white in Appearance >> Customize >> Typography (that might be necessary if the site background color is black) - the text is also white when editing pages making the text invisible.

**To reproduce**
1. Go to Appearance >> Customize >> Typography 
2. Set the text color and heading color to white and publish.
3. Head back to wp-admin >> Pages and open a random page
4. See the error

**Expected behavior**
I'd expect either the text to remain black on white background in page editor, or to see the background of the same color as the page so that the text is readable/editable.

**Screenshots**

<img width="293" alt="Screenshot 2019-07-11 at 16 23 21" src="https://user-images.githubusercontent.com/13084901/61054951-2980c580-a3f9-11e9-9dc6-2dcab2ebd590.png">

<img width="984" alt="Screenshot 2019-07-11 at 16 23 07" src="https://user-images.githubusercontent.com/13084901/61054960-2dace300-a3f9-11e9-8c58-c99b7051f800.png">

**Desktop (please complete the following information):**
 - OS: MacOs
 - Browser: Chrome
 - Version: 75.0.3770.100

**Additional context**
- WordPress 5.2.2

dphyled:
Another report of this issue in ticket 2300139-zen

fragglejack:
I have this problem also.

juliaamosova:
Thanks for raising this issue @grigorijurasov . I can reproduce it. 

Here is how the post looks at the front end:

![dark](https://user-images.githubusercontent.com/19143190/68949176-c09bc680-07b9-11ea-825e-a45ee66530cb.jpg)

And here is how it looks at the back end of you try to edit it using Gutenberg editor:

![light](https://user-images.githubusercontent.com/19143190/68949183-c42f4d80-07b9-11ea-817c-71464864b34b.jpg)

We should likely keep the text of the page / post black in the editor despite of the color it has at the front end.

ccwalburn:
Another user has reported this issue here: 2537490-zen

Robertght:
This is what I used a temporary workaround through the Code Snippets plugin:

add_action('admin_head', 'my_custom_fonts');

```
function my_custom_fonts() {
  echo '<style>
[...]

PR #1253, opened 2020-01-31
Add 'display=swap' to avoid invisible text

mrohnstock:
As documented [here](https://web.dev/font-display/?utm_source=lighthouse) it's pretty easy to avoid invisible texts by using (google-) webfonts.

Some more infos are available on the linked page.


Aljullu:
Hi @mrohnstock, thanks for your contribution. By default we trust the user agent to apply the best font-display strategy to websites. We know there are several approaches and some of them might serve different use cases, so by default we prefer to use the default one.

According to this, I will proceed closing this PR for now. If you think this is not the best resolution, feel free to open an issue to explain the problem from your point of view. You can also check some other issues like #1181 in case you feel they refer to your use cases.

PR #1256, opened 2020-02-11
Reset display of editor images

### LINKED ISSUE #1255: Empty Cart image is not centered
According to the designs, it should be centered:
imatge

Aljullu:
Fixes #1255. It centers the inner block image in the _Cart_ block.

imatge

This change will affect all images in the editor. I did some basic testing and everything was looking good to me, but we might consider deferring merging this PR until the next version of Storefront is released so we give more time for testing.

## Steps to test
1. Install the last version of Gutenberg and the dev version of WooCommerce Blocks.
2. Add a _Cart_ block and switch to the _Empty Cart_ view.
3. Verify the image is centered by default.

### DIFF
--- a/assets/css/base/gutenberg-editor.scss
+++ b/assets/css/base/gutenberg-editor.scss
@@ -247,7 +247,6 @@ figure {
 img {
 	height: auto; // Make sure images are scaled correctly.
 	max-width: 100%; // Adhere to container width.
-	display: block;
 }
 
 a {


assets/css/base/gutenberg-editor.scss
```diff
@@ -247,7 +247,6 @@ figure {
 img {
 	height: auto; // Make sure images are scaled correctly.
 	max-width: 100%; // Adhere to container width.
-	display: block;
```
senadir: Isn’t this change a bit wildcard and might affect all image blocks? would it make sense to scope to cart only?

senadir (changes requested):
yeah after testing this change does break other blocks like the Image block
image

senadir:
also, changes here might contradict with https://github.com/woocommerce/storefront/pull/1265

Aljullu:
> yeah after testing this change does break other blocks like the Image block

I'm unable to reproduce this. Is that with a specific version of Gutenberg or WP? :thinking: 

> Isn’t this change a bit wildcard and might affect all image blocks?

Yeah... I'm fine closing this PR and going with something more local. I still think, however, that forcing all images to `display: block;` will continuously bring issues like #1255 and #1263, because upstream projects rely on browser default styles and we are overriding those. So we will need to continuously add exceptions to that override.

> also, changes here might contradict with #1265

Right, I think #1265 wouldn't have been needed if we didn't set the `display` of all images to `block`.

haszari:
Marking this as blocked until we wrap up the discussion on #1297:  https://github.com/woocommerce/storefront/pull/1297#issuecomment-614403990

Aljullu:
Closing this PR since #1255 was fixed in WC Blocks.

PR #1257, opened 2020-02-12
fix untranslatable string ("or") in admin notice when WooCommerce plugin is deactivated

### LINKED ISSUE #1195: Untranslatable string
While preparing the release of Storefront 2.5.2, I noticed that the PR #1190 added a new string that is not wrapped in a call to one of the translation functions:

https://github.com/woocommerce/storefront/pull/1190/files#diff-63dca543345bcdd9dc4e4ecf64ce5025R113

I noticed this late in the release process and decided to continue with it instead of waiting for a fix that can be included in the next patch release.

cc @peterfabian

haszari:
Fixes #1195 

In Storefront, there's an admin notice displayed when WooCommerce is not active. One of the strings used for this was not translatable. This PR wraps it in a translation function so that it can be translated.

The string is the `or` in this notice:

Screen Shot 2020-02-12 at 1 15 40 PM

### How to test this PR

1. Clone this branch as a theme in your dev environment, and activate the theme.
2. Deactivate WooCommerce plugin.
3. View an admin page - should see notice. Confirm the notice is working correctly.

Note – if you've previously dismissed this notice you may need to re-enable it.

### DIFF
--- a/inc/admin/class-storefront-plugin-install.php
+++ b/inc/admin/class-storefront-plugin-install.php
@@ -110,7 +110,7 @@ if ( ! class_exists( 'Storefront_Plugin_Install' ) ) :
 				?>
 				<span class="plugin-card-<?php echo esc_attr( $plugin_slug ); ?>">
 					<a href="<?php echo esc_url( $button['url'] ); ?>" class="<?php echo esc_attr( $button['classes'] ); ?>" data-originaltext="<?php echo esc_attr( $button['message'] ); ?>" data-name="<?php echo esc_attr( $plugin_name ); ?>" data-slug="<?php echo esc_attr( $plugin_slug ); ?>" aria-label="<?php echo esc_attr( $button['message'] ); ?>"><?php echo esc_attr( $button['message'] ); ?></a>
-				</span> or
+				</span> <?php esc_html_e( 'or', 'storefront' ); ?>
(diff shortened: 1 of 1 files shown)


inc/admin/class-storefront-plugin-install.php
```diff
@@ -110,7 +110,7 @@ public static function install_plugin_button( $plugin_slug, $plugin_file, $plugi
 				?>
 				<span class="plugin-card-<?php echo esc_attr( $plugin_slug ); ?>">
 					<a href="<?php echo esc_url( $button['url'] ); ?>" class="<?php echo esc_attr( $button['classes'] ); ?>" data-originaltext="<?php echo esc_attr( $button['message'] ); ?>" data-name="<?php echo esc_attr( $plugin_name ); ?>" data-slug="<?php echo esc_attr( $plugin_slug ); ?>" aria-label="<?php echo esc_attr( $button['message'] ); ?>"><?php echo esc_attr( $button['message'] ); ?></a>
-				</span> or
+				</span> <?php esc_html_e( 'or', 'storefront' ); ?>
```
mikejolley: This needs context since "or" alone could mean many things. I'd suggest using https://developer.wordpress.org/reference/functions/_x/ and wrapping that in `echo esc_html(` to maintain the escaping also.

haszari: Agreed, this is not ideal in its current state – I'll take a look and get it tidied up.

PR #1198, opened 2019-08-20
Update all dev dependencies using `ncu -a`

rodrigoprimo:
This PR updates all NPM dev dependencies using `ncu -a` (npm-check-updates).

PR #1258, opened 2020-02-12
Update dev dependencies to latest minor version

haszari:
Updates all dependencies to latest minor version.

Also moves [browserslist](https://github.com/browserslist/browserslist) config out to conventional `.browserslistrc` file; this fixes a warning generated by postcss during build: 

```
  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config 
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.

  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist
```

This PR is intended as a replacement for #1198, but only updating to latest minor versions. #1198 updates all packages to latest, including major versions. We can look at more aggressive dependency updates in future; perhaps we should configure [Renovate](https://renovate.whitesourcesoftware.com) or similar to help with this.

### How to test this PR

- Clone this branch in your dev environment.
- Reinstall all dependencies – `npm ci`.
- Test Storefront styling, pages and core functionality – e.g. customizer, product pages, shop page, cart and checkout. Everything should work correctly (same as master/last release).

Note: I have compared newly-built `style.css` with the [last release](https://github.com/woocommerce/storefront/releases/tag/version%2F2.5.3) to confirm that [autoprefixer](https://github.com/postcss/autoprefixer) (which uses browserslist) is working correctly. Feel free to independently check this :)


.browserslistrc
```diff
@@ -0,0 +1,5 @@
+# Browsers that we support
+
+> 0.1%
+ie 8
+ie 9
```
Aljullu: Not blocking this PR, but at some point we should update this list. A good reference would be to copy or use [Gutenberg's](https://github.com/WordPress/gutenberg/tree/master/packages/browserslist-config).

haszari: Good point, perhaps we can pull in `@wordpress/browserslist-config` directly and customise if necessary.

PR #1264, opened 2020-02-20
force (reset) no underline on form required field indicator

### LINKED ISSUE #1262: Required checkout fields showing a dotted underline under the required indicator
## Describe the bug

When viewing the checkout page, the required fields show a red dotted underline under the `*` character:

image

The element which is used for this text is `<abbr>`, which some browser style by default with a dotted underline. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr

- [x] I have deactivated other plugins and themes and confirmed this bug occurs when only WooCommerce + Storefront theme are active.
- [x] I can reproduce this bug consistently using the steps below.

## To Reproduce
1. Go to https://themes.woocommerce.com/storefront/
2. Add a product to the cart and proceed to the checkout page
3. Notice the extra dotted underline on the required fields

## Expected behavior
I'd expect this style to be reset so it doesn't show an underline. The CSS for this element is already being altered here: https://github.com/woocommerce/storefront/blob/version/2.5.4/assets/css/woocommerce/woocommerce.scss#L1572-L1575
So it could get an extra line to reset the `text-decoration`

### Browser Environment
It will depend on the Browser which is being used, in my case I'm using Chrome on Ubuntu.

### WordPress Environment
Please provide relevant details of your WordPress setup and server environment.

<details>

```
WordPress 5.3.2
WooCommerce 3.9.2
Storefront 2.5.4
```
</details>

haszari:
Fixes #1262 

In WooCommerce, an asterisk is used to indicate required fields, e.g. checkout form. An `abbr` tag is used for this, and many browsers style this with a dotted underline. The dotted underline is not desirable :)

To fix this, this PR resets `.required` to have no underline - `text-decoration: none`.

Here's a screenshot of the checkout with the fix:

Screen Shot 2020-02-20 at 3 41 24 PM

However .. it seems to me this should be handled by WooCommerce core. It looks like WooCommerce has style overrides for various themes, but not Storefront, e.g. [twenty twenty](https://github.com/woocommerce/woocommerce/blob/master/assets/css/twenty-twenty.scss#L64). This could be fixed in Woo for Storefront also, or perhaps this underline should be reset for all themes by default. I'm curious why it's an `abbr` tag too.

I'll leave this PR open for now, and I'll report this issue to Woo core. cc @mikkamp

### DIFF
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -1571,6 +1571,7 @@ ul.order_details {
 
 .required {
 	border-bottom: 0 !important;
+	text-decoration: none;
 	color: $error;
 }
 


haszari:
Labelling this PR as blocked as I believe this should be fixed in WooCommerce core. We may still ship this fix in Storefront theme if needed.

haszari:
Logged this issue upstream: https://github.com/woocommerce/woocommerce/issues/25714

haszari:
Closing - this is logged upstream in WooCommerce Core.

PR #1265, opened 2020-02-20
Fix image alignment issue

### LINKED ISSUE #1263: [2.5.4] Images no longer properly wrapping on front end
## Describe the bug
Since 2.5.4, images that are set to wrap (i.e. left/right aligned) no longer wrap properly on the front end as they did before.

## To Reproduce

**Start with 2.5.3:**
1. Create a new page
2. Insert a paragraph block with content.
3. Insert an image block.
4. Left-align the image so that paragraph wraps around the image.
5. View front end. Everything is nicely wrapped as expected.

**Upgrade to 2.5.4:**

- You will note it remains correct on the backend.
- You will note the wrapping is not working correctly on the front end.

## Screenshots
These screenshots illustrate 2.5.4.

BACKEND (which displays properly after 2.5.4):
backend

FRONTEND (which does not display properly after 2.5.4):
frontend

## Expected behavior
The frontend should display as it did with 2.5.3, wrapping properly on the frontend.

FYI 1: I'm not using WooCommerce.
FYI 2: The same bug is present with Storefront, or with the Homestore child theme.

nielslange:
Fixes #1263 

<table>
<tr>
<td>Before (left aligned):
<br><br>

#1263-alignleft-before
</td>
<td>After (left aligned):
<br><br>

#1263-alignleft-after
</td>
</tr>
</table>

<table>
<tr>
<td>Before (right aligned):
<br><br>

#1263-alignright-before
</td>
<td>After (right aligned):
<br><br>

#1263-alignright-after
</td>
</tr>
</table>

### DIFF
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -305,7 +305,7 @@
 	}
 
 	// Image
-	.wp-block-image {
+	.wp-block-image:not( .block-editor-media-placeholder ) {
 		
 		&:not( .block-editor-media-placeholder ) {
 			display: inline;
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -306,6 +306,9 @@
 
 	// Image
 	.wp-block-image {
+		
+		display: inline;
+
 		figure {
 			margin-bottom: ms(2);
 		}
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -306,10 +306,7 @@
 
 	// Image
 	.wp-block-image:not( .block-editor-media-placeholder ) {
-		
-		&:not( .block-editor-media-placeholder ) {
-			display: inline;
-		}
+		display: inline;
 
 		figure {
 			margin-bottom: ms(2);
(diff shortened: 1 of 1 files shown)


assets/css/base/gutenberg-blocks.scss
```diff
@@ -306,6 +306,9 @@
 
 	// Image
 	.wp-block-image {
```
senadir: we should try to scope this to normal images only
```suggestion
	.wp-block-image:not( .block-editor-media-placeholder ) {
```

nielslange:
@senadir I've missed that I broke the image block flow and I missed that I could have accepted your proposed change directly. Instead, I went ahead and updated the PR manually and pushed the changes. 🤦‍♂️ The whole day, I'm facing a sore throats and headache. I should not create PRs in this condition. 😟

assets/css/base/gutenberg-blocks.scss
```diff
@@ -305,7 +305,12 @@
 	}
 
 	// Image
-	.wp-block-image {
+	.wp-block-image:not( .block-editor-media-placeholder ) {
+		
+		&:not( .block-editor-media-placeholder ) {
```
senadir: why do we have double classes here? Wouldn’t one be enough?

haszari: Agreed! I made this change and managed to push it to @nielslange branch (my plan was to set up a parallel branch in this repo so we can merge sooner).

nielslange: @senadir That was my bad. I manually added the change and then merged your suggestion as well, which lead to the duplicate code. Sorry for the confusion.

PR #1267, opened 2020-02-20
fix issue with top rated product not well aligned to the center

### LINKED ISSUE #1250: Product grid blocks: image is aligned to the left when there are few columns
**To Reproduce**
1. Add a _Top Rated Products_ block.
2. Set the number of columns to 1 or 2.
3. In the editor, notice the image is aligned to the left instead of being centered.

**Screenshots**
imatge

senadir:
fixes #1250 

it adds a well-scoped css to align the images

image

### DIFF
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -680,3 +680,11 @@
 		}
 	}
 }
+
+// Top rated products
+.wc-block-grid {
+	img {
+		display: block;
+		margin: 0 auto ms(3);
+	}
+}
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -347,11 +347,6 @@ ul.products,
 			font-size: ms(-1);
 		}
 
-		img {
-			display: block;
-			margin: 0 auto ms(3);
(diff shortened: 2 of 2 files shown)


assets/css/base/gutenberg-blocks.scss
```diff
@@ -680,3 +680,11 @@
 		}
 	}
 }
+
+// Top rated products
```
Aljullu: Actually that affects all product grid blocks (Top rated, Best sellers, Hand-picked, etc.), I would either remove this comment line or change it for something like:
```suggestion
// Product grid blocks
```

assets/css/base/gutenberg-blocks.scss
```diff
@@ -680,3 +680,11 @@
...
+.wc-block-grid {
+	img {
+		display: block;
```
Aljullu: I think we can remove this rule, at least it seems to be always overridden by `.editor-styles-wrapper img` for me.
```suggestion
```

assets/css/base/gutenberg-blocks.scss
senadir: Make sense, I first only applied to top-rated but then changed to grid products and didn’t update the comment

PR #1275, opened 2020-02-28
Small improvement to homepage.js

valdrinkoshi:
Avoids querying for the same elements multiple times.

### DIFF
--- a/assets/js/homepage.js
+++ b/assets/js/homepage.js
@@ -18,8 +18,9 @@
 			return;
 		}
 
-		for ( var i = 0; i < homepageContent.querySelectorAll( '.entry-title, .entry-content' ).length; i++ ) {
-			homepageContent.querySelectorAll( '.entry-title, .entry-content' )[ i ].classList.add( 'loaded' );
+		var entries = homepageContent.querySelectorAll( '.entry-title, .entry-content' );
+		for ( var i = 0, l = entries.length; i < l; i++ ) {
+			entries[ i ].classList.add( 'loaded' );
 		}
 
 		var siteMain         = document.querySelector( '.site-main' );
--- a/assets/js/homepage.js
+++ b/assets/js/homepage.js
@@ -19,7 +19,7 @@
 		}
 
 		var entries = homepageContent.querySelectorAll( '.entry-title, .entry-content' );
-		for ( var i = 0, l = entries.length; i < l; i++ ) {
+		for ( var i = 0, i < entries.length; i++ ) {
 			entries[ i ].classList.add( 'loaded' );
 		}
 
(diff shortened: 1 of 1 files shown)


assets/js/homepage.js
```diff
@@ -18,8 +18,9 @@
 			return;
 		}
 
-		for ( var i = 0; i < homepageContent.querySelectorAll( '.entry-title, .entry-content' ).length; i++ ) {
-			homepageContent.querySelectorAll( '.entry-title, .entry-content' )[ i ].classList.add( 'loaded' );
+		var entries = homepageContent.querySelectorAll( '.entry-title, .entry-content' );
+		for ( var i = 0, l = entries.length; i < l; i++ ) {
```
Aljullu, line 22: Being honest, I don't think there is a big performance gain assigning `entries.length` to `l`, but it does make the code a bit more verbose. Could we simplify it as so?
```suggestion
		for ( var i = 0; i < entries.length; i++ ) {
```

haszari:
> Thanks for the contribution @valdrinkoshi! Code looks safe and it worked fine in my testing.

Agreed, this is a nice tidy up, don't like seeing the query in the loop body - thanks :)

> will make it sense to add this to 2.5.6?

@Aljullu yes - added to milestone.

assets/js/homepage.js
haszari: I agree - also this is unusual style (multiple assign in the `for` statement). More idiomatic and no downside to just using `entries.length` directly as suggested above. Can you make this change please and then we'll merge :)

valdrinkoshi: Sounds good, done 👌

haszari: Thank you! Comment when you're done testing and I'll merge :)

valdrinkoshi: @haszari is that a task for me or @Aljullu?

haszari: Just confirming you've tested locally @valdrinkoshi :)

assets/js/homepage.js
```diff
@@ -18,8 +18,9 @@
 			return;
 		}
 
-		for ( var i = 0; i < homepageContent.querySelectorAll( '.entry-title, .entry-content' ).length; i++ ) {
-			homepageContent.querySelectorAll( '.entry-title, .entry-content' )[ i ].classList.add( 'loaded' );
+		var entries = homepageContent.querySelectorAll( '.entry-title, .entry-content' );
+		for ( var i = 0, i < entries.length; i++ ) {
```
Aljullu, line 22: Thanks for applying the feedback @valdrinkoshi! Just one last comment, I think there is a typo:
```suggestion
		for ( var i = 0; i < entries.length; i++ ) {
```

valdrinkoshi: Argh, great catch! Fixed 👌

PR #1284, opened 2020-03-03
document release process

haszari:
Adds a new document with details of how to release Storefront.

1. Read the [releases.md doc](https://github.com/woocommerce/storefront/blob/add/release-process-doc/docs/releases.md). Ensure this provides good guidance for how to release this theme, and doesn't include details that we'll need to update too often.

### Changelog

> Documented Storefront release process.


docs/releases.md
```diff
@@ -0,0 +1,80 @@
...
+### 4 – Launch release
+
+Now we get to release!
+
+- If necessary, build final `storefront.zip` with correct (non-RC) version number (repeat step 1).
+- Publish release on [GitHub](https://github.com/woocommerce/storefront/releases).
+  - Upload release zip.
+  - Use this format for tag: `version/1.2.3`.
+  - Paste changelog into “release details” field.
+- Upload to [WordPress.org](https://wordpress.org/themes/upload/). (You'll need access to the `Automattic` user account.)
```
nielslange: Can't we use [WordPress.org Plugin Deploy workflow](https://github.com/10up/action-wordpress-plugin-deploy) from 10up to deploy new releases from GitHub to WordPress.org without manually moving files?

haszari: That's a possibility, though it's not what we currently do. This PR is focused on documenting the existing process - we can look at improvements or changes as follow up.

Note that Storefront has a build process to generates `.css` files from `.scss` (etc), so the tagged version of the source code doesn't actually match the built theme. So it's important that we upload the built zip to WordPress.org. I'm not familiar with the 10up workflow but it may not apply directly to how this repo works.

nielslange: Sounds good. I'll add the plugin deployment workflow as a separate ticket that could be played in the future.

docs/releases.md
```diff
@@ -0,0 +1,80 @@
...
+__*Outcome*: A `storefront.zip` file is available to the community and other stakeholders for testing and feedback.__
+
+### 3 – Test & QA
+
+Note: all new code should have been fully tested during development. This test pass is to confirm that there are no major regressions or bugs, aka [“happy path”](https://en.wikipedia.org/wiki/Happy_path) or [smoke testing](http://softwaretestingfundamentals.com/smoke-testing/).
+
+- __Recommended__: Test in a clean environment, similar to a typical hosting environment. Avoid testing in your development environment.
+- __Recommended__: Test with a snapshot of data from a real store, with customers, products and reviews.
+- Test core WordPress and Woo flows.
+- Test core Storefront features.
```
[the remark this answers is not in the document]
haszari: I do think that's a good idea. Let's follow up on that separately - I'm thinking we need to flesh out that list of flows more. Best not to commit that to the repo until we have a plan for how we keep it up to date.

PR #1295, opened 2020-03-19
Fixed typo in css property.

yash-webkul:
Fixed typo issue in css property.

1. It will fix all the veritical alignments of the button, select, input and textarea to middle.

### Changelog

> Fixed typo issue in css property.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -846,7 +846,6 @@ select,
 textarea {
 	font-size: 100%; /* Corrects font size not being inherited in all browsers */
 	margin: 0; /* Addresses margins set differently in IE6/7, F3/4, S5, Chrome */
-	vertical-align: middle; /* Improves appearance and consistency in all browsers */
 }
 
 /* Override gutenberg styles. */
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -846,9 +846,7 @@ select,
 textarea {
 	font-size: 100%; /* Corrects font size not being inherited in all browsers */
 	margin: 0; /* Addresses margins set differently in IE6/7, F3/4, S5, Chrome */
-	vertical-align: baseline; /* Improves appearance and consistency in all browsers */
-	/* stylelint-disable */
-	*vertical-align: middle; /* Improves appearance and consistency in all browsers */
+	vertical-align: middle; /* Improves appearance and consistency in all browsers */
 	/* stylelint-enable */
 }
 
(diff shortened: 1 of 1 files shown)


assets/css/base/_base.scss
```diff
@@ -846,9 +846,7 @@ select,
 textarea {
 	font-size: 100%; /* Corrects font size not being inherited in all browsers */
 	margin: 0; /* Addresses margins set differently in IE6/7, F3/4, S5, Chrome */
-	vertical-align: baseline; /* Improves appearance and consistency in all browsers */
-	/* stylelint-disable */
-	*vertical-align: middle; /* Improves appearance and consistency in all browsers */
+	vertical-align: middle; /* Improves appearance and consistency in all browsers */
```
nielslange: According to https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align, `vertical-align: baseline;`  is the default value and when using `vertical-align: middle;` instead, I cannot see a change either. Wouldn't it make more sense to completely remove this definition from the current statement? 

@haszari  or @senadir Is one of you aware why the definition above had been added in the first place?

assets/css/base/_base.scss
```diff
@@ -846,9 +846,7 @@ select,
...
 	/* stylelint-enable */
```
nielslange: When removing `/* stylelint-disable */`, `/* stylelint-enable */` should also be removed.

nielslange:
Hello @yash-webkul 👋 Thanks a lot for your contribution! I reviewed your changes, added some comments and added @haszari and @senadir into this PR for their input on this change. 😉

yash-webkul:
Hello @nielslange Thanks for reviewing this.
Kindly check the both attached screenshots, first is with vertical-align: baseline and second one is with vertical-align: middle.
baseline
middle

See the difference between **Show Password** icon.
And sorry I forget to remove /* stylelint-enable */.
Thanks for mentioning it, I am removing it now.

nielslange:
Hello @yash-webkul, thanks for the screenshots. I cannot see a difference of the Show **password icon** though. I only see that the **Remember me** checkbox is aligned slightly different. This is a screenshot with both your screenshots above each other whereas the one on top has a transparency of 50%:

Screenshot 2020-03-21 at 10 57 42

May I ask what browser you are using? I removed the `vertical-align` statements completely and when using Chrome 80.0.3987.149 on macOS 10.15.3 and Safari 13.3.1 on iOS 13.3.1. I tend to removing the statement completely.

yash-webkul:
@nielslange yes you are right, i have removed the vertical-align property completely now.

nielslange:
Hello @yash-webkul 👋 I've already reviewed this PR and as mentioned in https://github.com/woocommerce/storefront/pull/1295#pullrequestreview-378894830 the change looks goo to me. That said, I'm not part of the WooCommerce team, that's why someone from the specific team need to do a second review to double-check the changes. Requesting multiple reviews, will not speed up the progress.

Issue #1285, opened 2020-03-05
Storefront Theme Overwrites Styles from Button Block


damiannep:
## Describe the bug
The following CSS in Storefront overrides settings made with Gutenberg:

```
/* Override gutenberg styles. */
.wp-block-button__link {
border-radius: 0 !important; }
```

## To Reproduce
1. Create a new page/post
2. Add a button block and set the radius to a non-zero value
3. Visit the live site
4. The button remains rectangular regardless of the radius set

## Expected behavior
Update the border-radius for buttons as set in the post/page

### WordPress Environment
WP Version: 5.3.2
WC Version: 3.9.3

Reported in 2758202-zen

damiannep:
Reported in 2758202-zen

nielslange:
Also reported in 19934108 - HC

haszari:
Reopening - PR for this issue was accidentally merged early - see #1322 (revert). 

See PR #1311 review for discussion on the issue and potential fix.

PR #1310, opened 2020-04-14
Remove underline from button blocks

Aljullu:
See related issue #1285

The underline was intended to be added to links that look like links.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Add a _Buttons_ block.
2. In the frontend, verify they don't have an underline.

### Changelog

> Buttons blocks no longer show an underline.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -455,13 +455,15 @@ body {
 		color: mix( $color_body, #222 );
 	}
 
-	a:not( .button ):not( .components-button ):not( .wp-block-button__link ) {
+	a {
 		color: $color_links;
 		text-decoration: underline;
 
 		&:hover {
 			text-decoration: none;
 		}
+
+		@include notUnderlinedWhenButton;
 	}
 }
 
@@ -783,12 +785,14 @@ table {
 	}
 
 	.entry-content {
-		a:not( .button ):not( .components-button ):not( .wp-block-button__link ) {
+		a {
 			text-decoration: underline;
 
 			&:hover {
 				text-decoration: none;
 			}
+
+			@include notUnderlinedWhenButton;
 		}
 	}
 
@@ -1545,8 +1549,9 @@ button.menu-toggle {
 			font-size: ms(2);
 		}
 
-		a:not( .button ):not( .components-button ):not( .wp-block-button__link ) {
+		a {
 			@include underlinedLink();
+			@include notUnderlinedWhenButton;
 		}
 	}
 }
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -463,7 +463,7 @@ body {
 			text-decoration: none;
 		}
 
-		@include notUnderlinedWhenButton;
+		@include removeButtonUnderline();
 	}
 }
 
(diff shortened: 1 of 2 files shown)


assets/css/base/_base.scss
```diff
@@ -1556,7 +1556,7 @@ button.menu-toggle {
 			font-size: ms(2);
 		}
 
-		a:not( .button ):not( .components-button ) {
+		a:not( .button ):not( .components-button ):not( .wp-block-button__link ) {
 			@include underlinedLink();
```
haszari, line 1560: These rules all look backwards to me – can you add a comment about the intention for each?

Aljullu:
I refactored this a bit to make code clearer and easier to reuse, it also decreases selector specificity. I did a round of testing and everything was looking good to me. But let me know if you catch something I missed.

> In my testing this fixes the button issue, but introduces a regression for regular text links (not a button) - these homepage links now have underline (it disappears on mouseover though grin) (featured categories, new products, also affects all products block):

Good catch! That was a specificity issue, should be fixed with the refactor.

> By the way, I there is an option for button component to render as a link – would be good to test that use case also if possible. That might need a custom block/plugin to test (not sure of the details).

Added an specific not selector for it: b2f9020db919843875944fec9f07f1637f52307c.

assets/css/sass/utils/_mixins.scss
```diff
@@ -32,6 +32,15 @@
 	}
 }
 
+// Remove underline if element has a button class.
+@mixin notUnderlinedWhenButton() {
```
haszari, line 36: I wonder if we can make this name clearer - starting with `not` always confuses me. Not a blocker but if you come up with something clearer before merge go for it :)

Does this remove underline from buttons? If so, we could call this `removeButtonUnderline` or similar.

assets/css/sass/utils/_mixins.scss
```diff
@@ -32,6 +32,15 @@
...
+	&.button,
+	&.components-button:not( .is-link ),
```
haszari, line 38: `stylelint` complained about this so I fixed it ☑️ b1c3ef31ef6a5f6abfb00912870016a9ad83cdff

```
assets/css/sass/utils/_mixins.scss
 38:26  ✖  Expected single space after "("    selector-pseudo-class-parentheses-space-inside
 38:33  ✖  Expected single space before ")"   selector-pseudo-class-parentheses-space-inside
```

assets/css/base/_base.scss
```diff
@@ -783,12 +785,14 @@ table {
...
 	.entry-content {
-		a:not( .button ):not( .components-button ) {
+		a {
 			text-decoration: underline;
 
 			&:hover {
 				text-decoration: none;
 			}
+
+			@include notUnderlinedWhenButton;
```
haszari, line 795: If we typically add the `()` to all mixins might be nice to do so here.

assets/css/sass/utils/_mixins.scss
Aljullu: Thank you! :tada:

assets/css/sass/utils/_mixins.scss
Aljullu: Yeah, much better name. Renamed in 97894ccfa67b3367a7535117b89afb7a1e978f6d.

assets/css/base/_base.scss
Aljullu: Done in 97894ccfa67b3367a7535117b89afb7a1e978f6d.

PR #1311, opened 2020-04-14
Do not force 0 border radius on buttons

### LINKED ISSUE #1285: Storefront Theme Overwrites Styles from Button Block

Aljullu:
Fixes #1285 

The _Buttons_ block allows specifying the border-radius, so we shouldn't force all buttons to have 0.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Add a _Buttons_ block and add three buttons: one without changing any attribute, one setting the border radius to 50, and the other one setting the border radius to 0.
2. Verify the border radius is honored in all of them.

### Changelog

> Buttons blocks now respect the border radius set by the user.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -852,11 +852,6 @@ textarea {
 	/* stylelint-enable */
 }
 
-/* Override gutenberg styles. */
-.wp-block-button__link {
-	border-radius: 0 !important;
-}
-
 button,
 input[type='button'],
 input[type='reset'],
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -171,7 +171,6 @@
 			text-shadow: none;
 			display: inline-block;
 			-webkit-appearance: none;
-			border-radius: 0;
 		}
 
 		&:not( .is-style-squared ) {


nielslange:
This would fix #1285

haszari:
Apologies @Aljullu  - I merged this PR by mistake. I've now reverted it (#1322) since we still need to figure out the right solution here. 

I've restored the branch but haven't opened a new PR. I'll leave that to you - the review comments above still apply.

Again, sorry for causing confusion here. 🤦

Aljullu:

> Defer this issue for 2.5.6 (ie no fix) - the existing opinionated behaviour ("Storefront has square buttons") is reasonable.

While we could defer fixing it after 2.5.6, I think at some point we will still need to fix. Currently, the border radius attribute of the _Buttons_ block feels broken in Storefront.

If we opted for an opinionated behavior, we should at least make it clear in the editor the border radius can't be changed. Otherwise it looks like a bug for the users: they can change the border radius in the editor but those changes aren't persisted in the frontend.

> Tweak the css so that Storefront homepage, all products etc are overridden to use square buttons, while allowing user to customise raw button block radius.

Yeah, that makes a lot of sense and would minimize the visual effect of this change. I created a PR that still enforces the border-radius to be 0, but removes the `!important` declaration, so button blocks can overwrite it. Let me know how it looks: #1323.

Issue #1287, opened 2020-03-05
Ensure compatibility with WordPress 5.4


GlennMartin1:
## WordPress 5.4 is schedule for release this month.
(per https://make.wordpress.org/core/5-4/)

## WordPress 5.4 introduces breaking changes:
https://make.wordpress.org/core/2020/03/03/wordpress-5-4-field-guide/
https://wptavern.com/preparing-for-wordpress-5-4-changes-theme-and-plugin-developers-should-know-about

Storefront should be reviewed and tested for WordPress 5.4, prior to its release.

nielslange:
**Things to check:**

- [ ] Calendar markup → https://make.wordpress.org/core/2020/02/12/changes-related-to-calendar-widget-markup-in-wordpress-5-4/
- [ ] New nav menu hooks → https://make.wordpress.org/core/2020/02/25/wordpress-5-4-introduces-new-hooks-to-add-custom-fields-to-menu-items/
- [ ] apply_shortcodes() alias function → https://make.wordpress.org/core/2020/02/13/wordpress-5-4-introduces-apply-shortcodes-as-an-alias-for-do-shortcode/

haszari:
Added conservative tags to this issue (enhancement, high priority). If there are specific breakages due to this we can upgrade to `Type: Bug` (etc) as needed.

Aljullu:
Thanks for the heads up @GlennMartin1 and the links @nielslange. I did a quick review and that's what I found out:
- Social icons block: seems to work ok without any changes. :heavy_check_mark: 
- Buttons block: seems to work ok without any changes. :heavy_check_mark: 
- Calendar markup: I created a PR (#1315) and an issue (#1314). Besides that, I think everything else is ok.
- New nav menu hooks: I don't think we need to do anything here.
- `apply_shortcodes()` as an alias for `do_shortcode()`: I'm not sure we want to do this change right now for backwards compatibility. Thoughts?
- `editor-` classes renamed to `block-editor-`: I created a PR (#1316).

If nothing else pops up, I think it should be fine to close this issue and open follow-ups if there is anything we missed.

GlennMartin1:
Thanks much! Any ETA on the next release?

nielslange:
> apply_shortcodes() as an alias for do_shortcode(): I'm not sure we want to do this change right now for backwards compatibility. Thoughts?

[...]

PR #1315, opened 2020-04-16
Wordpress 5.4 compatibility - Make calendar styles shared between the widget and the block

Aljullu:
Spun off from  #1287 - https://make.wordpress.org/core/2020/02/12/changes-related-to-calendar-widget-markup-in-wordpress-5-4/

Make the styles for the _Calendar_ widget shared with the _Calendar_ block. The only visual change is that the letter & numbers of the _Calendar_ block will now be centered.

### Screenshots
These screenshots show the _Calendar_ block on the left and the _Calendar_ widget on the right.
_Before:_
imatge

_After:_
imatge

1. Add a _Calendar_ widget in a sidebar and a _Calendar_ block in a post.
2. View them in the frontend and notice text is centered in the block.

### Changelog

> Text inside the Calendar block is now centered, in line with the Calendar widget.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -1512,6 +1512,7 @@ button.menu-toggle {
 	}
 }
 
+#wp-calendar, // Required by WP <=5.3 widget.
 .wp-calendar-table {
 	th,
 	td {
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -1512,14 +1512,11 @@ button.menu-toggle {
 	}
 }
 
-.widget-area,
-.footer-widgets {
-	#wp-calendar {
-		th,
-		td {
-			padding: 0.236em;
-			text-align: center;
-		}
+.wp-calendar-table {
+	th,
+	td {
+		padding: 0.236em;
+		text-align: center;
 	}
 }
 


haszari:
Updated the title and description to clarify the intentions/goal here, and bumped this up the priority in milestone :)

assets/css/base/_base.scss
```diff
@@ -1512,14 +1512,11 @@ button.menu-toggle {
...
 
-.widget-area,
-.footer-widgets {
-	#wp-calendar {
-		th,
-		td {
-			padding: 0.236em;
-			text-align: center;
-		}
+.wp-calendar-table {
```
haszari, line 1515: This breaks <5.4, since the `.wp-calendar-table` classes got added in 5.4. I downgraded WP and this is what the calendar block looks like (on front end):

Screen Shot 2020-04-20 at 11 57 22 AM

Perhaps we need to set up our shared calendar styling for `.wp-calendar-block, .wp-calendar-table` so it applies to both (even in older WordPress).

Up to you how to handle this – if this is complex and we have slightly different/degraded styles in older WordPress that's probably ok to ship.

Aljullu: Absolutely, I should have thought about that. I added back a selector targeting WP <=5.3 calendar widget (1b6e26f7b2370f8f605a64c4d3f2e9d0dc473016). I added a comment specifying that selector is only needed for WP <=5.3 in case at some point we drop support for old versions and we can remove it.

Aljullu:

> This is looking reasonable, though I'm not clear on the intentions - do we want to share the same styling for calendar widget and block so that we can fix the issues due to markup changes in 5.4?

Yes, I think that now that the block and the widget share the same markup, it makes sense to have the same styles for them.

Merging this PR for now. We can revert it later if needed.

PR #1326, opened 2020-04-22
Wrap long product names in grouped product, cart and checkout

### LINKED ISSUE #1259: Linked product info (in Grouped products) are cut off on product page for some viewport widths
Some grouped products do not show the prices for the individual products - they are 'cut off' on the screen.
86694777_10157058991763660_3221387572655685632_n
I can kind of fix it for affected products with:

.postid-2192 td, table th {
padding: 1em 0em !important;
text-align: left;
vertical-align: top;
}

but not ideal.

haszari:
Fixes #1259

This PR fixes layout issues with lengthy product names on grouped product page, cart, and checkout.

`overflow-wrap: anywhere` is used to instruct the browser to wrap preserving word boundaries and if necessary, break words. (Previously the checkout used `word-wrap: break-word` which is a [deprecated alias](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap).)

I've added this as a mixin, including `word-break: break-word;` which is needed for current Safari.

### Screenshots

Screen Shot 2020-04-22 at 4 37 54 PM

Screen Shot 2020-04-22 at 4 38 20 PM

Screen Shot 2020-04-22 at 4 54 31 PM

1. Add/edit some products and give them names with long words - e.g. long part number `PPXJOF123456789234-5678 Semi-circular O-ring, polycarbonate, perforated`.
2. Add these products to a grouped product.
3. Test product page - confirm layout works well on all screen sizes and browsers.
2. Add to cart.
1. Test cart and checkout and confirm layout works correctly.

Prizes should always be visible and text/content should not overlap.

Please test in a variety of browsers – I haven't tested in IE yet so would appreciate if someone can test there :) 

### Changelog

> Fix – Product page, cart, and checkout layout fixes when product names are long.

### DIFF
--- a/assets/css/sass/utils/_mixins.scss
+++ b/assets/css/sass/utils/_mixins.scss
@@ -66,5 +66,6 @@
 	overflow-wrap: anywhere;
 	// Safari supports word-break.
 	word-break: break-word;
-	// (We may also need word-wrap for IE.)
+	// We also need word-wrap and for IE.
+	-ms-word-break: break-all;
 }
--- a/assets/css/sass/utils/_mixins.scss
+++ b/assets/css/sass/utils/_mixins.scss
@@ -58,3 +58,13 @@
 	border-radius: 0;
 	box-shadow: inset 0 -1px 0 rgba( #000, 0.3 );
 }
+
+@mixin wrapBreakWord {
+	// https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
+	// https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
+	// This is the current standard, works in most browsers.
+	overflow-wrap: anywhere;
+	// Safari supports word-break.
+	word-break: break-word;
+	// (We may also need word-wrap for IE.)
+}
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -572,7 +572,7 @@ ul.products,
 
 			table.woocommerce-grouped-product-list {
 				.woocommerce-grouped-product-list-item__label {
-					word-break: break-word;
+					@include wrapBreakWord();
 				}
 
 				.woocommerce-grouped-product-list-item__quantity {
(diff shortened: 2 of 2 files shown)


assets/css/sass/utils/_mixins.scss
```diff
@@ -49,3 +49,13 @@
 	border-radius: 0;
 	box-shadow: inset 0 -1px 0 rgba( #000, 0.3 );
 }
+
+@mixin wrapBreakWord {
+	// https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
+	// https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
+	// This is the current standard, works in most browsers.
+	overflow-wrap: anywhere;
```
Aljullu, line 57:
```suggestion
	overflow-wrap: anywhere;
	// IE11 doesn't support overflow-wrap neither word-break: break-word, so we fallback to work-break: break-all.
	word-break: break-all;
```

haszari: Agreed - I'll test in IE today and use that or `word-wrap` alias, which looks like it might work in IE ([was previously an MS extension](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)). Incidentally that what was used previously for the `table.woocommerce-checkout-review-order-table` case, so probably not a dangerous change :)

haszari: `-ms-word-break` did the trick. 21fa05f49cf3b377db3c2fe3f20076f49f66725b

haszari:
Added `-ms-word-break`, inspired by [ever-reliable CSS Tricks](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/).

Tested in IE and re-tested in Chrome, Safari, Firefox on mac - merging.

Screen Shot 2020-04-23 at 10 27 08 AM

Screen Shot 2020-04-23 at 10 28 49 AM
Screen Shot 2020-04-23 at 10 29 13 AM

Issue #1333, opened 2020-04-24
Storefront icons.css typo line 4439


Marmio:
Maybe old news, but.... for those who come across a conflict with fontawesome text-file icon on header menu. Woocommerce/Storefront comes with a typo.  Check this:

file: [icons.scss](https://github.com/woocommerce/storefront/blob/master/assets/css/base/icons.scss#L114)

```
.main-navgation ul li a::before {
  display: inline-block;
  font-style: normal;
  font-variant: normal;
  font-weight: normal;
  line-height: 1;
  font-family: 'Font Awesome 5 Free';
  font-weight: 900;
  line-height: inherit;
  vertical-align: baseline;
  content: "\f15c";
  margin-right: 0.5407911001em; }
```

should be (my guess): 
.main-navigation (i/o  'navgation')

haszari:
Thanks for reporting this @Marmio - good spotting. I've edited the description for clarity.

haszari:
Closing this in favour of #1347.

Usually I'd close the newer issue but in this case #1347 has a recommended plan of action. Thanks for reporting @Marmio  :)

Issue #1319, opened 2020-04-20
fix broken icons in comment list/form (or remove broken css if problematic)


haszari:
## Describe the bug
There's some [css in `icons.css` which adds icons for comment author, "edit comment" pencil etc](https://github.com/woocommerce/storefront/blob/master/assets/css/base/icons.scss#L34-L63).

This is no longer working - it looks like the parent selectors (e.g `#comments commentlist`) have changed at some point.

Perhaps we can remove this css. Or if these icons are important (or this is a recent regression) we can fix them up.

## Screenshots
This is current behaviour, no icons:

<img width="832" alt="Screen Shot 2020-04-20 at 3 57 52 PM" src="https://user-images.githubusercontent.com/4167300/79712723-c4f77d00-831f-11ea-96bc-83544c25f473.png">

If I hack `commentlist` back in we get some nice icons:

<img width="875" alt="Screen Shot 2020-04-20 at 3 59 25 PM" src="https://user-images.githubusercontent.com/4167300/79712779-ec4e4a00-831f-11ea-9617-2a807598cabe.png">

(Discovered when reviewing #1279.)

nielslange:
Due to aesthetic and usability reasons, I'd rather keep the icons and correct the CSS code, so that the icons become visible again.

haszari:
> keep the icons and correct the CSS code, so that the icons become visible again

@nielslange I'm happy with that as long as it's safe/low risk and we have time to test it :)

I'm a little curious how long this has been broken. If it's multiple releases then perhaps it's more risky to bring it back, as it might feel like an unexpected visual change to some stores. The risk is probably pretty low though (at a guess).

haszari:
Adding to 2.5.7, hopefully we can tidy this up safely.

gglobalstep:
This issue is observed to be **Open** on WooCommerce 4.2 RC 1
**Environment:**
Windows
OS: 10
Browser: Chrome(Version 81.0.4044.138 (64-bit))

MAC 
OS: Mojave v10.14.6
Browser: Safari(v13.0.5)

haszari:
Thanks for testing this @gglobalstep :)

Can you confirm the Storefront version where you saw the issue? This issue is fixed in Storefront 2.5.7 RC 1.

Please ensure you are testing with Storefront 2.5.7 RC 1 by following the steps here - https://github.com/woocommerce/storefront/wiki/Release-testing-instructions:-Storefront-2.5.7

[...]

PR #1334, opened 2020-04-24
correct typo in icons.scss

GlennMartin1:
correct typo in icons.scss

navigtion > navigation

### DIFF
--- a/assets/css/base/icons.scss
+++ b/assets/css/base/icons.scss
@@ -111,7 +111,7 @@ input[type='submit'],
 	}
 }
 
-.main-navigtion {
+.main-navigation {
 	ul {
 		li {
 			a {


GlennMartin1:
I'm a novice at this, but I think https://github.com/woocommerce/storefront/issues/1333 is **not** referring to the the same code as this PR.

Aljullu:
Why do you think so @GlennMartin1? Your PR is modifying `icons.scss`, after build (running `npm run build` in a terminal) it becomes `icons.css`. Line numbers change between them but that's to be expected. I checked line 4439 of `icons.css` and it matches the selector you modified in `icons.scss`, so I think they are the same. Does that make sense?

GlennMartin1:
Ok, very well; I don't understand the build process. The reason for my confusion is this:

- This pull request targets this misspelling: `.main-navigtion`.
- But issue https://github.com/woocommerce/storefront/issues/1333 targets this misspelling `.main-navgation`.

Those are different misspellings.

Aljullu:
Aaah, good catch @GlennMartin1. I think that's a typo in the issue description: `icons.css` L4439 is `.main-navigtion ul li a::before {` for me

haszari:
> It seems that this has been broken for more than 3 years now, so I think most stores expect the icon not to appear. In my opinion, instead of fixing this we should plainly remove that block of CSS (L114-L126). How does it sound?

> Pinging @haszari for his opinion as well.

I agree with @Aljullu, if these icons have not been used for three years, it's probably not a great idea to add them back unless there's a request/report to do so. Looking at that screenshot it's not clear to me that these are useful - they just show a "document" icon for all nav options, which doesn't seem right.

Note there's a similar issue with icons in comments (#1319). In that case I'm leaning towards fixing it because the icons provide a clear benefit, and the fix won't change something so critical as the main nav/header.

So I'm thinking we tidy this css up - remove the broken rule for these icons, rather than fix it. Note we don't need to consider this for 2.6.1 :)

haszari:
Closing this PR - as discussed above, we've decided not to pursue the fix.

Logged a new issue for removing the stale css, after appropriate investigation / testing: https://github.com/woocommerce/storefront/issues/1347

PR #1338, opened 2020-04-28
add basic testing flows doc (very basic)

haszari:
This PR adds a basic checklist of flows supported in Storefront, to aid testing. 

This is not an exhaustive list, and some of the items may need to be clarified.

Also the categories/headings are draft - keen for feedback on how to simplify this and make it useful and easy to keep up to date.

In general - keen for feedback :)

Is there anything missing - what does Storefront do/support that we should test?

Is anything not clear?

How should we organise this?

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -35,7 +35,7 @@ This document lists the features and flows supported in Storefront. This is not 
 - [ ] On smaller screens, a sticky footer provides easy access to important site functions (for example, shopping cart).
 
 ### Testing tips
-- If you're not seeing the onboarding notice, delete the  `storefront_nux_dismissed` option to bring it back.
+- If you're not seeing the onboarding notice, delete the  `storefront_nux_dismissed` [option](https://codex.wordpress.org/Options_API) to bring it back. For example, with [WP CLI](https://developer.wordpress.org/cli/commands/option/): `wp option delete storefront_nux_dismissed`.
 
 ## Customization
 - [ ] Merchant can use custom colors for site typography / text.
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -2,54 +2,60 @@
 
 This document lists the features and flows supported in Storefront. This is not an exhaustive list.
 
-## WooCommerce core
-- [ ] single product page
-- [ ] shop page / product catalog / product list
-- [ ] search for products
-- [ ] sale products
-- [ ] add to cart, view cart, manage cart contents
-- [ ] use coupon
-- [ ] checkout
-- [ ] log in as customer, view account info
-- [ ] customer can view previous orders
-- [ ] visitor/customer can review products
-- [ ] visitors can view reviews
-- [ ] woo widgets
+## WooCommerce - core eCommerce features
+- [ ] Explore product catalogue on shop page and category pages.
+- [ ] Explore specific products on single product page.
+- [ ] Search for products using keywords.
+- [ ] Identify products which are currently on sale.
+- [ ] Add products to cart, view cart, manage cart contents.
+- [ ] Checkout and complete purchase.
+- [ ] Use coupons in cart/checkout for a discount or free shipping.
+- [ ] Sign up for a customer account.
+- [ ] Log in as customer, view account info, manage account, change password.
+- [ ] Customer can view previous orders.
+- [ ] Visitors and customers can review and rate products.
+- [ ] Visitors can view reviews & ratings to get "social proof" about product quality/performance.
+- [ ] Merchant can customise eCommerce related features in pages and layout areas via WooCommerce widgets.
 
-## WooCommerce blocks
-- [ ] show products
-- [ ] explore / search products
-- [ ] show reviews
-- [ ] all products / grid blocks
+## WooCommerce - editing and publishing content
+- [ ] Merchant can build pages and posts showing specific products.
+- [ ] Merchant can build pages and posts for exploring / searching product catalogue.
+- [ ] Merchant can use reviews on pages and posts.
+- [ ] Merchants and build pages and posts showing a grid of products that the customer can filter/search.
+
+### Testing tips 
+- This category also covers traditional shortcodes.
 
 ## Storefront features
-- [ ] onboarding / welcome notice / create homepage (unset storefront_nux_dismissed)
-- [ ] homepage template
-- [ ] full width page template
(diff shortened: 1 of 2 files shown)


docs/testing.md
```diff
@@ -0,0 +1,55 @@
...
+- [ ] search for products
+- [ ] sale products
+- [ ] add to cart, view cart, manage cart contents
+- [ ] use coupon
+- [ ] checkout
+- [ ] log in as customer, view account info
+- [ ] customer can view previous orders
+- [ ] visitor/customer can review products
+- [ ] visitors can view reviews
+- [ ] woo widgets
```
Aljullu, line 17: Ideally I think most bullet points should have at least one verb that specifies what needs to be tested. In some cases it might be a bit ambiguous otherwise (do I need to check whether I can add Woo Widgets, whether they appear or whether there have style regressions?).

docs/testing.md
```diff
@@ -0,0 +1,55 @@
...
+- [ ] explore / search products
+- [ ] show reviews
+- [ ] all products / grid blocks
+
+## Storefront features
+- [ ] onboarding / welcome notice / create homepage (unset storefront_nux_dismissed)
+- [ ] homepage template
+- [ ] full width page template
+- [ ] sticky add-to-cart header
+- [ ] block editor adapts to full/sidebar width as appropriate (editor.js)
```
Aljullu, line 30: What does `editor.js` mean in this line?

docs/testing.md
haszari: I totally agree – and this helps to soften the checklist and focus on the real need, rather than how it's implemented. I'll do another pass on this.

docs/testing.md
haszari: This is an interesting case – I don't fully understand the rationale for the feature, and want to avoid detailing all the specifics of how it works. I added the filename for the code because there might be other aspects to the feature that we need to cover. We might not need this item specifically - it might be covered by `editor styles match front end` for example. 

The goal here is to have a flow item(s) that captures the benefits for merchants. 

A related consideration is how critical/important each flow is. Perhaps we need a column/rating for criticality (e.g. `essential | useful | low priority`.

haszari:
Thanks @Aljullu - I've reworked things quite a bit thanks to your prompting, and added some new ones too.

The other change I've made is to add a `Testing tips` heading under each category. This allows us to document specific details and tricks/testing instructions where needed, while keeping the flows focused on the high level. The idea is that some of the stuff under `Testing tips` might change over time, but the flows we support are much more constant/stable.

Can you give it one more read through before we merge? We'll need to keep an eye on these and iterate each time we release, but I'd also like to get as much feedback incorporated in this first version.

docs/testing.md
Aljullu: I think it's much clearer now. Thanks for expanding it!

nielslange:
Great work, @haszari! I'd probably structure and write the test cases slightly different, e.g. 

**WooCommerce » shopping**

**Store owner**

- can look up products on shop page and product category pages
- can look up product details on product page
- can search for products
-  can do ...

**Customer**

- can look up cart page and checkout page
- can look up my account page
-  can do ...

**WooCommerce » publishing**

**Store owner**

-  can do ...

**Customer**

-  can do ...

This naming convention allows both reading and implementing the test cases very easy.

docs/testing.md
```diff
@@ -0,0 +1,61 @@
+# Testing Storefront
```
senadir, line 1: Shameless plug but I think this list needs some [automated todos.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/docs/testing/cart-checkout/coupons.md)

haszari: Feel free to submit a PR! Though I would caution against relying on these items – they are intended to spark good holistic testing of the product, rather than prescribe all the required things to tick off to guarantee quality :)

docs/testing.md
```diff
@@ -0,0 +1,61 @@
...
+- [ ] Search for products using keywords.
+- [ ] Identify products which are currently on sale.
+- [ ] Add products to cart, view cart, manage cart contents.
+- [ ] Checkout and complete purchase.
+- [ ] Use coupons in cart/checkout for a discount or free shipping.
+- [ ] Sign up for a customer account.
+- [ ] Log in as customer, view account info, manage account, change password.
+- [ ] Customer can view previous orders.
+- [ ] Visitors and customers can review and rate products.
+- [ ] Visitors can view reviews & ratings to get "social proof" about product quality/performance.
```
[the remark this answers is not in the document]
haszari: Good point, I'll simplify it, I think the purpose of reviews doesn't need to be explained here :)

256b6c2b3e0fca5dc099f27bedca87e60a2f1bd2

docs/testing.md
```diff
@@ -0,0 +1,61 @@
...
+## Storefront features
+- [ ] Storefront shows an onboarding / welcome notice which guides user to set up Storefront and create a homepage.
+- [ ] Homepage template showcases product catalogue.
+- [ ] Merchant can author pages that use the full page width.
+- [ ] Merchant can enable a sticky "add to cart" header on product pages so CTA button doesn't get lost when scrolling product page.
+- [ ] Visitor can view and zoom product images on product pages, and use swipe gestures to walk through images on mobile devices.
+- [ ] On smaller screens, a sticky footer provides easy access to important site functions (for example, shopping cart).
+
+### Testing tips
+- If you're not seeing the onboarding notice, delete the  `storefront_nux_dismissed` option to bring it back.
```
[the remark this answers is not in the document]
haszari: I'll link this to WordPress wp options docs, don't want to get too prescriptive in how to test. I've added some more info here: f485f8054b5c6c11378cc523edd3fbe68efa432e

haszari:
> I'd probably structure and write the test cases slightly different, e.g.

Thanks for the review @nielslange. I'm trying to avoid making this look too much like a specific list of all things to check. For this reason I've not organised it by user role (though some user roles are mentioned to help illustrate the flow/need); many of the flows can and should be tested by a variety of roles, and I don't want to prescribe all those combinations. For example "search for products" is primarily a visitor/shopper/customer flow, but also relevant for admins/merchants.

That said, I'm definitely keen to iterate on how we structure these to more clearly define what Storefront offers and how it works. Feel free to submit PRs to improve :)

PR #1368, opened 2020-06-04
Remove woocommerce-legacy.scss

Aljullu:
[Storefront 2.5.7 supports WC >=4.0](https://docs.woocommerce.com/document/storefront-requirements-version-support/), so there is no reason to keep stylesheets for legacy versions of WC.

Verify there are no regressions in styles.

### Changelog

> Fix – Removed `woocommerce-legacy.css` stylesheets for unsupported versions of WooCommerce (< 4.0). #1368


haszari:
Note: I'm testing this with WooCommerce Blocks deactivated because of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2646. In any case we'll need to give this a thorough test before release.

inc/woocommerce/class-storefront-woocommerce.php
```diff
@@ -150,11 +150,6 @@ public function woocommerce_scripts() {
 			if ( ! class_exists( 'Storefront_Sticky_Add_to_Cart' ) && is_product() ) {
 				wp_register_script( 'storefront-sticky-add-to-cart', get_template_directory_uri() . '/assets/js/sticky-add-to-cart' . $suffix . '.js', array(), $storefront_version, true );
 			}
-
-			if ( defined( 'WC_VERSION' ) && version_compare( WC_VERSION, '3.3', '<' ) ) {
```
haszari, line 154: We should probably search the code for other uses of `WC_VERSION` - maybe there's other stuff we can tidy up.

Aljullu: Sure thing! I found this one while working on #1369, but there might be others.

haszari: #1373 💥

PR #1369, opened 2020-06-04
Set storefront-style and storefront-icons as dependencies of storefront-woocommerce-style

### LINKED ISSUE #1299: Styles are not enqueued with explicit dependencies
## Overview

If a child theme enqueues a stylesheet with a defined Storefront dependency, like so:
```php
wp_enqueue_style(
    'child-theme-style',
    get_stylesheet_directory_uri() .'/assets/styles/main.css',
    ['storefront-woocommerce-style']
);
```
it will break the order of the style tags of Storefront itself, which can (will) lead to broken styling _(in this particular example, `storefront-woocommerce` stylesheet will get enqueued before `storefront-icons` which will break the breadcrumbs.)_.

##  To Reproduce

1. Start with a clean Storefront
2. Add and activate a rudimentary child theme (style and functions.php)
3. Enqueue a child style in a manner similar to the above code (declare a dependency on `storefront-woocommerce-style`)
4. Observe the bug with the breadcrumbs

#### Example of broken breadcrumbs
image

#### Workaround

This could be mitigated by supplying a priority higher than `30` to `wp_enqueue_scripts` (for this particular issue), but is, arguably, not the most easily maintainable way to go about this.

#### Fix
Explicitly declare dependencies of styles in Storefront, rather than rely on the implicit order of enqueue, so that WP makes the right choice when enqueueing, even with dependencies coming from "outside".

Aljullu:
Fixes #1299.

### Screenshots
_Before:_
imatge
_(notice the breadcrumbs are broken)_

_After:_
imatge

In order to test this PR, you will need to create a dummy Storefront child theme as follows:

`functions.php`:
```PHP
<?php
add_action( 'wp_enqueue_scripts', 'enqueue_styles' );

function enqueue_styles() {
	wp_enqueue_style( 'storefront-child-style-2', get_stylesheet_directory_uri() . '/style-2.css', ['storefront-child-style', 'storefront-woocommerce-style'] );
}
?>
```

`style.css`:
```CSS
/*
 Theme Name:   Storefront Child
 Template:     storefront
*/
body {
	border: 5px solid pink;
}
```
`style-2.css`:
```CSS
body {
	border-top-color: blue;
}
```

1. Select that theme in WP-Admin.
2. Verify styles are correctly loaded and breadcrumbs don't break.

### Explanation

Notice, before this PR, there was a work-around to achieve the same result as the `After` screenshot above that consisted on increasing the priority value of `wp_enqueue_scripts` in `functions.php`:
```PHP
<?php
add_action( 'wp_enqueue_scripts', 'enqueue_styles', 40 );

function enqueue_styles() {
	wp_enqueue_style( 'storefront-child-style-2', get_stylesheet_directory_uri() . '/style-2.css' );
}
?>
```

So what this PR does is that instead of needed the work-around, child themes can define `storefront-woocommerce-style` as a style dependency and doing that won't change the order the styles are loaded so the layout doesn't break.

### Changelog

> Fix – Stylesheet `storefront-woocommerce-style` has `storefront-style` and `storefront-icons` as explicit dependencies, so child themes can define `storefront-woocommerce-style` as a dependency without the risk of breaking the stylesheets order.

### DIFF
--- a/inc/woocommerce/class-storefront-woocommerce.php
+++ b/inc/woocommerce/class-storefront-woocommerce.php
@@ -139,7 +139,7 @@ if ( ! class_exists( 'Storefront_WooCommerce' ) ) :
 
 			$suffix = ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ? '' : '.min';
 
-			wp_enqueue_style( 'storefront-woocommerce-style', get_template_directory_uri() . '/assets/css/woocommerce/woocommerce.css', array(), $storefront_version );
+			wp_enqueue_style( 'storefront-woocommerce-style', get_template_directory_uri() . '/assets/css/woocommerce/woocommerce.css', array( 'storefront-style', 'storefront-icons' ), $storefront_version );
 			wp_style_add_data( 'storefront-woocommerce-style', 'rtl', 'replace' );
 
 			wp_register_script( 'storefront-header-cart', get_template_directory_uri() . '/assets/js/woocommerce/header-cart' . $suffix . '.js', array(), $storefront_version, true );


Aljullu:
Thanks for reviewing and expanding the explanation @haszari!

> @Aljullu how did you determine the missing dependencies (`storefront-style`, `storefront-icons`)? Is it possible there are other dependencies for our other stylesheet handles that we should register too?

I checked the order styles are loaded when there are no child themes and compared it when a child theme has a stylesheet with `storefront-woocommerce-style` as a dependency. From there, it was mostly trial-and-error, `storefront-icons` needed to be loaded before  `storefront-woocommerce-style` to avoid the breadcrumbs breaking.

I couldn't see any issues if `storefront-style` was not declared as a dependency, but the default SF order was that `storefront-style` was loaded first, so it made sense to also define it as a dependency of `storefront-woocommerce-style`.

In the other hand, it seemed `storefront-fonts` didn't need to be loaded in any specific order, that's why I didn't include it in the dependencies.

PR #1394, opened 2020-06-18
Basic e2e test setup using wp-env, jest and puppeteer [in progress]

### LINKED ISSUE #1330: Implement e2e testing
Similar to https://github.com/woocommerce/woocommerce/tree/master/tests/e2e-tests#how-to-run-an-individual-test, having e2e tests in place might help to reduce bugs while implementing new features and updating existing features.

haszari:
Related #1330 - this is the first step, a few follow up tasks mentioned here.

The goal here is to get a basic e2e setup so that we can incrementally add tests for front-end & admin dashboard flows. This will provide an early warning when PRs impact existing flows (catch regressions) and another automated way to test releases. Also as we build more tests this will provide a good reference for what flows are supported in Storefront.

This PR adds basic config for the following:

- A [wp-env](https://developer.wordpress.org/block-editor/packages/packages-env/) test container (server) environment, for running tests against. Could be used as a standard/default dev setup for the repo too potentially.
- Jest & Puppeteer and related dependencies. This is the combination used by other projects, notably WooCommerce core. Jest is the test runner and Puppeteer allows tests to target UI flows in a browser - aka end-to-end (e2e) tests.
- A very basic test to confirm that the `Built with Storefront` footer text is present 😁

#### What needs doing?
- ~~The store url is hard-coded - need to configure this. @senadir can you advise how to do this? I see in WooCommerce Blocks there's a `default.json` config file with env vars, but I don't know how to tell Jest/puppeteer about such a file.~~ Update: now using [jest `globals`](https://jestjs.io/docs/en/configuration#globals-object), this works well. However I'm still interested to hear other / better approaches.
- Docs for how to run tests and contribute new tests, and how it all works.
- Explore and clarify our needs re WooCommerce and WordPress test utils. As we start implementing tests for real flows we will want to leverage a lib for navigating admin dashboard and other common flows not implemented by us. None of this is in place yet - TBD if this is a blocker for this PR.
- Longer term, I'd love to be able to run core WooCommerce and WordPress flows as part of our test suite. Examples: commenting on a post, or add-to-cart and complete purchase. I suspect that we'll be able to write these tests by using utils libs, am curious if we can share more of these flows tests. 

1. Get storefront built: clone storefront repo, `npm install`, `composer install`, `npm run build`.
2. Start wp-env environment: `wp-env start` from repo folder. 
4. When it's complete you should be able to see woo+storefront site at http://localhost:8888 - note that this is using the default ports. If you are running another env you'll need to take it down (`wp-env stop`). 
3. Run the tests: `npm run e2e` from repo folder.

Should output happy test result:

Screen Shot 2020-06-18 at 4 56 13 PM

Note about ports - I'm keen to set up Storefront on custom ports so it doesn't clash with other projects using the default, but I hit an issue with that - https://github.com/WordPress/gutenberg/issues/23259

### Changelog

> Dev - Add support for end-to-end tests using Jest & Puppeteer. #1394

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,0 +1,9 @@
+{
+    "core": null,
+    "plugins": [
+        "https://downloads.wordpress.org/plugin/woocommerce.latest-stable.zip"
+    ],
+    "themes": [
+        "./"
+    ]
+}
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,9 +1,9 @@
 {
-    "core": null,
-    "plugins": [
-        "https://downloads.wordpress.org/plugin/woocommerce.latest-stable.zip"
-    ],
-    "themes": [
-        "./"
-    ]
+  "core": null,
+  "plugins": [
+    "https://downloads.wordpress.org/plugin/woocommerce.latest-stable.zip"
+  ],
+  "themes": ["./"],
+  "port": 8801,
+  "testsPort": 8802
 }
--- a/package.json
+++ b/package.json
@@ -16,7 +16,9 @@
   },
   "jest": {
     "preset": "jest-puppeteer",
-    "setupFilesAfterEnv": ["expect-puppeteer"],
+    "setupFilesAfterEnv": [
+      "expect-puppeteer"
+    ],
     "globals": {
       "STORE_URL": "http://localhost:8889"
     }
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
       "expect-puppeteer"
     ],
     "globals": {
-      "STORE_URL": "http://localhost:8889"
+      "STORE_URL": "http://localhost:8801"
     }
   },
   "license": "GPL-2.0+",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,12 @@
     "build": "grunt",
     "css": "grunt css",
     "watch": "grunt watch",
-    "deploy": "grunt deploy"
+    "deploy": "grunt deploy",
+    "e2e": "jest"
+  },
+  "jest": {
+    "preset": "jest-puppeteer",
+    "setupFilesAfterEnv": ["expect-puppeteer"]
   },
   "license": "GPL-2.0+",
   "main": "Gruntfile.js",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,10 @@
   },
   "jest": {
     "preset": "jest-puppeteer",
-    "setupFilesAfterEnv": ["expect-puppeteer"]
+    "setupFilesAfterEnv": ["expect-puppeteer"],
+    "globals": {
+      "STORE_URL": "http://localhost:8889"
+    }
   },
   "license": "GPL-2.0+",
   "main": "Gruntfile.js",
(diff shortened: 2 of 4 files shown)


haszari:
This is in progress but I'm keen for feedback/ideas - @senadir  (or @nerrad if you have time) please take a look and comment with ideas for next steps or issues I might not be aware of. I've taken a from-scratch approach to the config, rather than paste lots of stuff over from Woo or Woo blocks - so there's probably important config still needed.

Super basic right now - the main blocker is moving the site url out to config. Keen for your thoughts on the minimum tests we need to know the right infra is in place. Example: what flows can we test if we set up woo test utils?

.wp-env.json
```diff
@@ -0,0 +1,9 @@
+{
+  "core": null,
```
nerrad, line 2: You can leave this property out and it will default to the latest stable WordPress version.

package.json
```diff
@@ -11,13 +11,25 @@
...
+    "e2e": "jest",
+    "wp-env": "wp-env"
+  },
+  "jest": {
+    "preset": "jest-puppeteer",
+    "setupFilesAfterEnv": [
+      "expect-puppeteer"
+    ],
+    "globals": {
+      "STORE_URL": "http://localhost:8801"
```
senadir, line 24: You probably want to expose this as an env variable instead of a global, globals are bits of code that should be accessed anywhere (`wp`, `wc`, ...).

haszari: Agreed - this is a bit temporary. 

For now I'm happy with this, as env vars are more global. But I agree we will likely iterate on this and it may well end up in an env var if needed.

haszari:
Thanks for fixing those up @nerrad ! Having `wp-env` as a local dependency is preferable anyway.

> I see you went with a very barebones installation of jest and puppeteer. That's okay, but there's some configuration and utils that would be handy pulling from the related @wordpress packages that could eventually be handy. Definitely something that could be iterated on as needed I suppose.

Agreed - and we'll likely need this when we start implementing more realistic tests. This is a very basic starting point :)

I'll leave the issue #1330 open after this merges, and will add specific follow ups based on the discussion in this PR (and the follow ups in the description).

.wp-env.json
haszari: Left this as is, no danger, and we may well tidy up in a future iteration.

haszari:
Added a bunch of follow ups to https://github.com/woocommerce/storefront/issues/1330 - feel free to add comments/edit description on that issue if I've missed anything.

package.json
nerrad: Using a `.env` file is handy.

haszari: Agreed! Does wp-env support these too? I guess maybe it does if it's docker-compose under the hood ..

haszari:
Just saw the redirect-to-8888 issue again and figured out what's happening. Google Chrome is [caching a previous redirect](https://superuser.com/questions/304589/how-can-i-make-chrome-stop-caching-redirects/869739) (not sure where the redirect came from, maybe an old wp-env instance).

Fix is to clear the browser cache, specifically `Cached images and files` :)

package.json
nerrad: > Does wp-env support these too?

`wp-env` does not support `.env` files. But something like [`dotenv`](https://www.npmjs.com/package/dotenv) makes it fairly trivial to support `.env` files in any other scripts we do (hence the suggestion).

PR #1419, opened 2020-07-12
Merge duplicate CSS selectors

Ismail-elkorchi:
The `_base.scss` file contains some duplicated selectors with the same properties in one case, and different ones for the others. This PR seeks to simplify the concerned selectors by merging them.

### Changelog

> Fix - Merge duplicate CSS selectors.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -233,15 +233,11 @@ figure {
 	margin: 0;
 }
 
-table {
-	margin: 0 0 ms(2);
-	width: 100%;
-}
-
 img {
 	height: auto; // Make sure images are scaled correctly.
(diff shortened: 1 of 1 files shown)


Ismail-elkorchi:
@nerrad Can you please elaborate on why you removed this PR from the `2.5.9` milestone?

nerrad:
Thanks for the ping, removal wasn't intentional, I was cleaning up some other pulls. 2.6.0 is the next release coming out.

Ismail-elkorchi:
Ah ! That makes sense. Thank you @nerrad for your careful explanation. I wasn't familiar with the release process here, so I wanted to check in to see if there is any problem with the PR.

PR #1422, opened 2020-07-14
Fix incorrect starter content for HomePage.

### LINKED ISSUE #1423: homepage start content contains invalid (broken) WooCommerce Blocks markup
## Describe the bug
The Storefront homepage starter content uses WooCommerce Blocks for some page elements. The markup used includes shortcodes as a fallback, and this syntax is no longer valid. 

Because of this, when the home page is opened in the block editor (aka Gutenberg), there are validation errors (`This block contains unexpected or invalid content.`) for various blocks - "New In" (latest products), "Fan Favorites" (top rated), etc.

Screen Shot 2020-07-15 at 12 06 55 PM

## To Reproduce
1. Start with a clean empty woo site, with Storefront `Let's go` admin onboarding notice displaying. 
   - If this notice is not shown, delete the `storefront_nux_dismissed` option, e.g. `wp option delete storefront_nux_dismissed`. 
2. Click `Let's go` to trigger the onboarding/starter content.
3. Customizer launches. Click `Publish` to complete onboarding.
4. Open admin pages list `Dashboard > Pages > All Pages`.
5. Edit `Homepage`. 

Note the validation errors in the blocks.

## Expected behavior
No validation errors in starter content - all blocks should work correctly and merchant should be able to edit settings easily.

Shortcodes only used when there is no appropriate block.

When blocks are used, there's no shortcode fallback contained within the block markup.

nerrad:
Fixes #1423

In testing, some folks surfaced a problem where the created Homepage via the starting notice for Storefront included invalid blocks. This surfaces when trying to edit the homepage. The blocks themselves render on the frontend, but the block structure invalidates when run through the block editor validation.

The blocks are recoverable, but this is a poor experience for users wanting to edit their homepage.

It looks like the starter content originally included shortcodes as a fallback when blocks were not present in the environment the theme is installed in. However it's been a while since that was the case and we can remove that extraneous fallback.

## To Reproduce
While on master:

Create a home page using this wizard:

Image 2020-07-14 at 2 34 07 PM

(You can trigger this via a fresh install of Storefront, or by manipulating the `storefront_nux_dismissed` and storefront_nux_fresh_site` options in your database).

When editing the home page you'll see this:

Image 2020-07-14 at 2 38 05 PM

## To Test

- Try creating a homepage using this branch.
- The homepage blocks should render as expected and when editing the homepage there should be no block validation errors.

### DIFF
--- a/inc/nux/class-storefront-nux-starter-content.php
+++ b/inc/nux/class-storefront-nux-starter-content.php
@@ -1048,9 +1048,7 @@ if ( ! class_exists( 'Storefront_NUX_Starter_Content' ) ) :
 				<h2 style="text-align:center">' . __( 'New In', 'storefront' ) . '</h2>
 				<!-- /wp:heading -->
 
-				<!-- wp:woocommerce/product-new {"columns":4} -->
-				<div class="wp-block-woocommerce-product-new">[products limit="4" columns="4" orderby="date" order="DESC"]</div>
-				<!-- /wp:woocommerce/product-new -->
+				<!-- wp:woocommerce/product-new {"columns":4} /-->
 
 				{{handpicked-products}}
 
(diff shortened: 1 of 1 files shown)


haszari:
Thanks for jumping on this @nerrad! I logged an issue for clarity - #1423.

haszari:
> I wonder if we should put anything in place to catch issues like this in future? If you have any ideas, feel free to tweak the wiki flows doc (or consider for blocks release process perhaps).

Update: I added `Ensure homepage has no errors on front end or editor.` which should be a good reminder :)

PR #1424, opened 2020-07-15
Skip styling All Products Product Title if we have a font size or a color from the editor

senadir:
In the newest version of  WooCommerce Blocks, users can customize All Product's Product Title block, this customization is via color or font size, but those changes can contradict with any style provided by a theme, such as Storefront.
This PR skips styling for that block if a user has already selected some styling, this is done via `has-text-color` and `has-font-size` classname.

See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2860 for more details about this behavior.

## Testing

In order to test this, you will need to have WooCommerce Blocks 3.0.0 enabled in your website.

- Create a page.
- Insert `All Products` Block in it.
- Click the "Edit" button of that block.
image
- Now select the Product Title block.
- In the sidebar, you can change the font size for that block, select a predefined font size for the list, not a custom font size value, as this issue is only present with predefined sizes.
- After changing the color, click "Done"
image
- Save the page and view it.
- Check if Storefront honors the font size you selected.

### Changelog

> Respect user-selected Color and Font size values for All Products' Product Title Block.

### DIFF
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -345,9 +345,15 @@ ul.products,
 		.woocommerce-loop-product__title,
 		.wc-block-grid__product-title,
 		.wc-block-grid__product-title > a {
-			font-size: 1rem;
 			font-weight: 400;
 			margin-bottom: ms(-3);
+		}
+		h2, // @todo Remove when WooCommerce 2.8 is live
+		h3, // @todo Remove when WooCommerce 2.8 is live
+		.woocommerce-loop-product__title,
+		.wc-block-grid__product-title,
+		.wc-block-grid__product-title > a:not(.has-font-size) {
+			font-size: 1rem;
 		}
 
 		.star-rating {
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -352,7 +352,7 @@ ul.products,
 		h3, // @todo Remove when WooCommerce 2.8 is live
 		.woocommerce-loop-product__title,
 		.wc-block-grid__product-title,
-		.wc-block-grid__product-title > a:not(.has-font-size) {
+		.wc-block-grid__product-title > a:not( .has-font-size ) {
 			font-size: 1rem;
 		}
 


nerrad:
Hi Nadir,

I realize we have a bit of context from the slack conversation but can you fill out a bit more details in the pull description about  why this is needed and the impact of the change. Also the testing instructions could include some more details. What specifically should be tested?

I know I could make a guess and some assumptions based on the pull but some more details will help. This is helpful for a few reasons:

- People browsing pulls have a better idea of what is happening in the pull without having to be aware of the context from discussions outside of the pull (especially useful for theme developers that may be watching what happens in Storefront to help guide what they do in their own themes).
- Preparing release notes that include changelogs and testing instructions.
- Future maintainers (which may be our future selves) troubleshooting issues that might be related to a pull. If we came across this pull in debugging something, would there be enough info to help understand why it was created without guessing?

assets/css/woocommerce/woocommerce.scss
```diff
@@ -345,10 +345,16 @@ ul.products,
 		.woocommerce-loop-product__title,
 		.wc-block-grid__product-title,
 		.wc-block-grid__product-title > a {
-			font-size: 1rem;
 			font-weight: 400;
 			margin-bottom: ms(-3);
 		}
+		h2, // @todo Remove when WooCommerce 2.8 is live
+		h3, // @todo Remove when WooCommerce 2.8 is live
```
nerrad, line 352: Should this be the following?

```suggestion
		h2, // @todo Remove when WooCommerce Blocks 2.8 is live
		h3, // @todo Remove when WooCommerce Blocks 2.8 is live
```

If so, since WooCommerce Blocks 2.8 has been out for a while, is this pull even needed anymore? Or do we still need to explicitly remove the `font-size: 1rem` as done above for the title selectors?

senadir: I'm not fully sure, I copied that from https://github.com/woocommerce/storefront/blob/f9a328f5105ea0efbd3ce7e9dc67d471997d54a8/assets/css/woocommerce/woocommerce.scss#L343-L349
That code is 4 years old, so I will go wager it was targeting WooCommerce 2.8, will delete it.

senadir:
I updated the description, thanks for letting me know @nerrad, I sometimes forget myself when doing small PRs.

assets/css/woocommerce/woocommerce.scss
Aljullu: I think that if we remove the `h2` and `h3` selectors, we will need to add `.woocommerce-loop-category__title`, otherwise Categories titles would have a different style from Products.

So I would keep this as-is in this PR and open an issue or a follow-up PR with the goal to replace `h2` and `h3` with  `.woocommerce-loop-category__title`.

---

For reference:

* `h2` and `h3` selectors were added in 19a4c5f46787856e767b632fafee035798107b51
* The `.woocommerce-loop-product__title` class name was introduced in https://github.com/woocommerce/woocommerce/pull/11583.

Aljullu (approved):
Works great. I added a comment about the `h2` and `h3` selectors, but I think it's good to go as-is. :ship:

PR #1472, opened 2020-09-19
👌 IMPROVE: Adjust pagination thumbnail size (#1429)

### LINKED ISSUE #1429: Add an option to set single product pagination image size for page optimization
The function `storefront_single_product_pagination()` doesn't allow any control over the image size, and so the default `woocommerce_thumbnail` size is used.

The `woocommerce_thumbnail` size is too big for this context, and hence triggering "Serve scaled images" notification on Google PageSpeed.

nielslange:
Fixes #1429

Adjust the pagination thumbnail size. Initially, the pagination thumbnails were using the image size 324px by 324px. As the pagination thumbnails are displayed with an image size of 90px by 90px. I introduced the image size `pagination_thumbnails` with an image size of 180px by 180px. I doubled the expected size so that the image looks [sharp on retina screens](https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/image-size-and-resolution/).

1. Install and set up the [WooCommerce](https://wordpress.org/plugins/woocommerce/) plugin.
2. Install and set up the [Storefront](https://woocommerce.com/storefront/?aff=10486&cid=1131038) theme.
3. Head over to a product page and check the image size of the pagination thumbnails within the inspector.

⚠️  If you have an existing WooCommerce test site, you need to regenerate the thumbnails to see this effect. You can achieve this via:

- [WP-CLI command `wp media regenerate`](https://developer.wordpress.org/cli/commands/media/regenerate/)
- [Regenerate Thumbnails plugin](https://wordpress.org/plugins/regenerate-thumbnails/)
- [reGenerate Thumbnails Advanced plugin](https://wordpress.org/plugins/regenerate-thumbnails-advanced/)

### Changelog

> Tweak - Adjust the pagination thumbnail size. #1429

### DIFF
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -71,6 +71,11 @@ if ( ! class_exists( 'Storefront' ) ) :
 			 */
 			add_theme_support( 'post-thumbnails' );
 
+			/*
+			 * Enable support for pagination thumbnails.
+			 */
+			add_image_size( 'pagination-thumbnails', 180, 180, true );
+
 			/**
 			 * Enable support for site logo.
 			 */
--- a/inc/woocommerce/storefront-woocommerce-template-functions.php
+++ b/inc/woocommerce/storefront-woocommerce-template-functions.php
@@ -774,14 +774,14 @@ if ( ! function_exists( 'storefront_single_product_pagination' ) ) {
 		<nav class="storefront-product-pagination" aria-label="<?php esc_attr_e( 'More products', 'storefront' ); ?>">
 			<?php if ( $previous_product ) : ?>
 				<a href="<?php echo esc_url( $previous_product->get_permalink() ); ?>" rel="prev">
-					<?php echo wp_kses_post( $previous_product->get_image() ); ?>
+					<?php echo wp_kses_post( $previous_product->get_image( 'pagination-thumbnails' ) ); ?>
 					<span class="storefront-product-pagination__title"><?php echo wp_kses_post( $previous_product->get_name() ); ?></span>
 				</a>
 			<?php endif; ?>
 
 			<?php if ( $next_product ) : ?>
 				<a href="<?php echo esc_url( $next_product->get_permalink() ); ?>" rel="next">
-					<?php echo wp_kses_post( $next_product->get_image() ); ?>
+					<?php echo wp_kses_post( $next_product->get_image( 'pagination-thumbnails' ) ); ?>
 					<span class="storefront-product-pagination__title"><?php echo wp_kses_post( $next_product->get_name() ); ?></span>
 				</a>
 			<?php endif; ?>


haszari:
Thanks for the PR @nielslange !

This looks like it could be a potential fix for the reported issue.

I've responded on the issue asking for a bit more detail about the impact of the issue, or the specific problem noted by PageSpeed Insights.

If the rendered size is 90px (180px retina) and the thumbnail size is 324px, I wouldn't necessarily assume that needs a new thumbnail size. As a rule of thumb, 324px is less than double 180px, so it's in the ballpark (and these are already somewhat small). With good image compression settings this might not be a major performance issue (probably depends on various factors on individual site). 

I'm also curious about the potential impact of adding a new thumbnail size vs. overriding the size of [`woocommerce_thumbnail`](https://docs.woocommerce.com/document/image-sizes-theme-developers/). Will this generate another copy of each product image on the server? I'm not an expert on WordPress image sizes/thumbnails so I'm keen to fully understand all the alternatives here and ensure we ship something that works well for all Storefront users.

nielslange:
> If the rendered size is 90px (180px retina) and the thumbnail size is 324px, I wouldn't necessarily assume that needs a new thumbnail size. As a rule of thumb, 324px is less than double 180px, so it's in the ballpark (and these are already somewhat small). With good image compression settings this might not be a major performance issue. The actual impact likely depends on various factors on each individual site.

In my opinion, images should be loaded in the correct size due to performance reasons. This is especially important in eCommerce as a long loading time lead to a higher dropdout rate and therefore less profit for the customer.

> I'm also curious about the potential impact of adding a new thumbnail size vs. overriding the size of [`woocommerce_thumbnail`](https://docs.woocommerce.com/document/image-sizes-theme-developers/). Will this generate another copy of each product image on the server? I'm not an expert on WordPress image sizes/thumbnails so I'm keen to fully understand all the alternatives here and ensure we ship something that works well for all Storefront users.

In my PR, I haven't overwritten existing functionality, but have introduces a new image size specifically for the pagination images. 

> Another thing to consider - can this be improved/optimised on sites that need the optimisation, without changing Storefront's base behaviour for all sites. Is there a workaround, e.g. in a child theme – or does this need to be fixed in Storefront core?

Of course, this functionality could be placed inside a child theme. In fact, everything can be placed in a child theme. However, if a customer does not use a child theme, the pagination thumbnails are still loaded in a higher resolution than needed. In addition, adding this functionality to a child theme would mean that it has to be added into every single child theme that is using the pagination feature. In my opinion, a child theme should only contain functionality that is not part of the parent theme itself. In this specific case, I see more disadvantages that advantages by moving this functionality into a child theme.

Of course, these are my personal thoughts on this case.

haszari:
> Of course, this functionality could be placed inside a child theme. In fact, everything can be placed in a child theme.

If this can be implemented in a child theme then I think it's best to recommend that until we have more evidence that it's causing a performance issue, and/or impacting sales on a significant number of stores. 

Adding a new thumbnail size is a big change for a small performance enhancement, so on balance I don't think we should do this by default in Storefront core. Storefront should be a lean, stable base which works well for the majority of sites; customization (via child theme) is a good option if sites need to optimise this specific aspect.

cc @jconroy - Keen for your thoughts or a second opinion here.

findingsimple:
👋 howdy @haszari  @nielslange 

At face value, I've never had too many issues in the past with adding images sizes for my own projects.

> Will this generate another copy of each product image on the server?

My understanding is that new media will automatically start generating the new sizes (in addition to existing sizes etc.)

Existing media won't generate this size until specifically told to regenerate (wp-cli, plugin etc.)

> images should be loaded in the correct size due to performance reasons. 

I agree with that as best practice

---

Looking specifically at this changeset (feel free to correct me), If customers don't regenerate their media thumbnails after this change (I suspect quite a few won't) - instead of the current 324px by 324px thumbnail size they will end up with the full image instead which could be a lot bigger (depending on their site of course).

For example i just tested the change and I start seeing the full 800 x 600 product images ( "pagination-thumbnails" size doesn't exist so it falls back to the full size). I feel like this potentially goes against the intent of optimizing if our existing less tech savvy users (who simply update their theme) start serving up much bigger images instead of just the slightly too big ones they currently have.

From that point of view, I agree with @haszari that we probably shouldn't rush into making this change.

> The function storefront_single_product_pagination() doesn't allow any control over the image size,

I'd personally err on simply adding a filter here to allow the default woo thumbnail size to be easily overridden in the pagination function (for those that want to optimse without overriding the whole `storefront_single_product_pagination()` function - which is already possible by the looks) and keep an eye out for more reports/concerns about this.

nielslange:
Thanks for your thoughts, @jconroy. I like the idea of adding a filter to `storefront_single_product_pagination()` to allow optimizing the pagination thumbnails for those who want to. I'll update this PR accordingly the next days.

asafm7:
> > Of course, this functionality could be placed inside a child theme. In fact, everything can be placed in a child theme.
> 
> If this can be implemented in a child theme then I think it's best to recommend that until we have more evidence that it's causing a performance issue, and/or impacting sales on a significant number of stores.
> 
> Adding a new thumbnail size is a big change for a small performance enhancement, so on balance I don't think we should do this by default in Storefront core. Storefront should be a lean, stable base which works well for the majority of sites; customization (via child theme) is a good option if sites need to optimise this specific aspect.
> 
> cc @jconroy - Keen for your thoughts or a second opinion here.

@nielslange  @haszari @jconroy 

Actually, this **_cannot_** be fixed with a child theme, without overriding the whole `storefront_single_product_pagination()` function. Overriding this function makes things prone to break on theme updates. This is a crucial point.

Maybe a middle ground will be adding the ability to add width and height attributes to the img tag (controlled by filters) so tools, like Jetpack Site Accelerator can fetch the correct size dynamcly, without creating a new hardcoded image size, and more image versions on the server.

budzanowski:
Closing. It looks like this is not too important in general and at the same, the consensus is that we should use a filter - probably in a fresh PR.

PR #1481, opened 2020-09-24
clarify release process: release from trunk, test as needed for release risk / complexity

haszari:
This PR has some tweaks to the release process, some based on @senadir feedback (thank you!).

- Storefront releases from default branch, and rarely uses release branch. 
- Version numbers and all other changes for the release should be pushed to master before releasing a build (RC or final).
- Testing should be appropriate for the release. I.e. small, low risk releases don't need as much testing; large, complex releases with many changes should have broader testing.
- Patch releases are rare/uncommon. 
- Version number scheme is based on WordPress core.


docs/releases.md
```diff
@@ -56,16 +62,17 @@ Testing documentation is available in the [wiki](https://github.com/woocommerce/
...
 
-In addition, please [review recent dependency updates](https://github.com/woocommerce/storefront/pulls?q=is%3Apr+author%3Aapp%2Frenovate+is%3Aclosed) and ensure that these changes are well tested.
+Perform testing as appropriate for the release. Use your judgement to ensure effective efficient testing – e.g. for small releases, 1 tester (you) x ~1 hour. For larger releases, allow more time and request assistance from team, testing guild as necessary.
+
+At a minimum, test the following:
+
+- Spot-check a cross-section of key flows - e.g. timeboxed exploratory testing.
+- Confirm recent fixes and features (this release, previous release) are working correctly and haven’t regressed.
+
+__If there are blocking issues or major regressions, the process stops here!__ (So they can be fixed.)
```
senadir, line 75: I like the reduced testing here, feels appropriate.

senadir (approved):
I like the new changes, I still have one complaint, it's still not obvious that you need to start with Pca54o-28-p2 before jumping to this document, therefor forgetting several key steps that need to be done.
Can we include that document as part of the handoff (the current release lead will ping the next one and include a link to that document).

haszari:
Thanks for the feedback @senadir !

> I like the new changes, I still have one complaint, it's still not obvious that you need to start with Pca54o-28-p2 before jumping to this document, therefor forgetting several key steps that need to be done.

Ah I see – I agree with you, though I think the team docs and process should take precedence over the repo docs. This repo and release process is "contained" within the (larger) team rotation; linking back out to private team docs could be confusing. As you've highlighted, within our team, we should refer to the rotation/team docs for as the primary source of truth. 

The intention with this release process doc is that anyone reading it has a clear idea of what happens in a Storefront release. I think it meets that goal.

PR #1502, opened 2020-10-16
Update Autoprefixer and replace grunt-postcss

Aljullu:
This PR supersedes #1464.

* It updates autoprefixer to the latest release.
* It also installs postcss 8.1, which is required by the last autoprefixer version.
* It also replaces `grunt-postcss` (whose [repo](https://github.com/nDmitry/grunt-postcss) has been archived) with `@lodder/grunt-postcss`, which is compatible with postcss 8.

In order to test this PR, I compared all CSS files generated in `trunk` and the ones generated in this PR. The only difference was a [`-moz-placeholder`](https://caniuse.com/css-placeholder-shown) prefix removed in `style.css`. According to our browser support policy, it makes sense that it has been removed.

### How to test the changes in this Pull Request:

1. Make sure to clean all packages and install the new ones (`npm ci`).
2. Run `npm run css` and navigate through the store to verify there are no styling regressions.

### Changelog

> Update – Update Autoprefixer to the latest version

PR #1503, opened 2020-10-16
Install cssnano to minify CSS files

Aljullu:
While working on #1502, I noticed Storefront CSS files were not minified. This PR installs cssnano (which is also [used in WC Blocks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/postcss.config.js#L4)) in order to minify CSS files.

Some filesize wins:

| File | `trunk` | This branch |
| :--- | ---: | ---: |
| `icons.css` | 103,2 kB | 83,8 kB |
| `woocommerce.css` | 76,8kB | 60,8 kB |
| `style.css` | 69,5 kB | 40,5 kB |
| `gutenberg-blocks.css` | 47,3 kB | 43,0 kB |
| `gutenberg-editor.css` | 5,9 kB | 4,7 kB |

Run `npm build` and verify there are no style regressions in the store.

### Changelog

> Performance – Added CSS minification.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,14 +4,17 @@ module.exports = function( grunt ) {
 
 	var sass = require( 'node-sass' );
 
+	var processors = [ require( 'autoprefixer' )() ];
+	if ( grunt.option('env') === 'production' ) {
+		processors.push( require( 'cssnano' )() );
+	}
+
 	grunt.initConfig({
 
 		// Autoprefixer.
 		postcss: {
 			options: {
-				processors: [
-					require( 'autoprefixer' )()
-				]
+				processors: processors
 			},
 			dist: {
 				src: [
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -4,14 +4,17 @@ module.exports = function( grunt ) {
 
 	var sass = require( 'node-sass' );
 
+	var processors = [ require( 'autoprefixer' )() ];
+	if ( grunt.option('env') === 'production' ) {
```
senadir, line 8: Would this make the most sense? compared to the regular `NODE_ENV` we use in most projects?
This is a loosely held opinion.

senadir (approved):
This tests fine for me and I think it's safe to merge.

Gruntfile.js
Aljullu: I don't think `NODE_ENV` is available inside grunt. At least, I couldn't get it working.

PR #1508, opened 2020-10-20
Add accent color to WC Blocks price slider

Aljullu:
This PR adds styles to the price slider of the Filter Products by Price block. It requires the changes in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3300, which decrease the selector specificity of the default styles.

#### Customizer
Peek 2020-10-20 18-25

#### Child theme
(Screenshots with [Bistro](https://woocommerce.com/products/bistro/))

_Before:_
Price filter screenshot

_After:_
Price filter screenshot with custom styles

1. Create a page with the price filter and the All Products block.
2. In the customizer, change the accent color to something different from purple.
3. Verify the price slider has the new accent color.
4. Test a Storefront child theme too to ensure the accent color of the child theme is honored as well.

### Changelog

> Enhancement – Added styles to the Filter Products by Price block, so the accent color is used in the slider.

### DIFF
--- a/inc/customizer/class-storefront-customizer.php
+++ b/inc/customizer/class-storefront-customizer.php
@@ -870,6 +870,11 @@ if ( ! class_exists( 'Storefront_Customizer' ) ) :
 				color: ' . $storefront_theme_mods['hero_text_color'] . ';
 			}
 
+			.wc-block-components-price-slider__range-input-progress,
+			.rtl .wc-block-components-price-slider__range-input-progress {
+				--range-color: ' . $storefront_theme_mods['accent_color'] . ';
+			}
+
 			.wc-block-components-button:not(.is-link) {
 				background-color: ' . $storefront_theme_mods['button_alt_background_color'] . ';
 				color: ' . $storefront_theme_mods['button_alt_text_color'] . ';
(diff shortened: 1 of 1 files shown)


senadir (approved):
Those changes are not being honoured on the editor, this can be addressed in a follow-up.

On initially loading this PR, I see this on frontend
Screen Shot 2020-11-02 at 1 14 13 PM
In the editor, I see this:
Screen Shot 2020-11-02 at 1 15 35 PM

When changing a color, the result is reflected on the frontend, but not the editor.

On trunk, I can see the new purple on both editor and frontend, so I don't know why is the frontend using old purple in this PR.

senadir:
I missed this step

> In WC Blocks directory check out the branch update/styles-source: git fetch && git checkout update/styles-source.

senadir:
After doing that step, I'm still seeing this issue in the editor (color not being respected).

Aljullu:
> Those changes are not being honoured on the editor, this can be addressed in a follow-up.

Good catch @senadir! I fixed this in 07f58462ec54ccf9d6c5ff00060f611655d9de68. The issue was affecting other blocks like Cart and Checkout (they were not honoring the button color in the editor). I moved the CSS of all blocks so it should be fixed in all.

> After doing that step, I'm still seeing this issue in the editor (color not being respected).

Ah, sorry, I just updated the PR to remove that step. It's no longer needed since that PR was merged into WC Blocks.

Aljullu:
I will proceed merging this since it has already been approved and this way it gets in `trunk` for tomorrow's RC. We will still have one week of testing just in case something unexpected happens.

PR #1525, opened 2020-11-02
Remove editor dynamic width feature

### LINKED ISSUE #1512: WordPress 5.6 compatibility – editor width doesn't change based on full-width/default template

Aljullu:
Closes #1512.

As discussed in #1520, we want to remove the feature to adapt the editor width to the selected template.

1. Create a page and switch between the 'Full width' template and the 'Default' template.
2. Verify there are no visible changes in the editor, but when previewing the page in the frontend, the correct template is applied.
3. Since this PR includes some changes to the built process, it might be good to test other JS features to ensure there are no regressions. For example, the sticky add to cart bar that can be added from the Customizer (under the WooCommerce > Product Page section) and appears in (not-external) product pages when you scroll down:
imatge

Affected [flows & features](https://github.com/woocommerce/storefront/wiki/Testing-Storefront:-flows-and-features):

- Block editor shows a realistic preview of the front end styling.

### Changelog

> Removal – Removed the feature to adapt the editor width to the selected template.


Aljullu:
I didn't set a specific milestone because I don't think that needs to block 2.9.0 release.

package.json
```diff
@@ -34,7 +34,6 @@
   "devDependencies": {
     "@babel/core": "7.12.3",
```
haszari: Do we need `babel` dependencies any more (`@babel/core grunt-babel`)? I'm assuming that editor.js was the only file using ES6/babel-powered JS syntax - is that correct? Looks like babel is (mostly) removed from Gruntfile (in this PR). 

There are two options:

1. Keep babel installed and run the task, even if it's not needed. This is "future proof", we have babel running and can update existing code to use ES6 when fixing bugs.
2. Completely remove babel entirely, since it's not strictly needed right now. 

@Aljullu Up to you which approach we take here. From a quick look at the gruntfile, I think we need to remove `grunt.loadNpmTasks( 'grunt-babel' );` as well (if you decide on option 2).

haszari (approved):
Great to see the code base getting simpler (more robust). I tested editing a page and changing templates, and also tested most JS functionality (sticky add-to-cart, product pagination, mobile footer, mobile menu) – everything worked well in my testing (Chrome / Safari). 

Overall this is looking splendid 👌, but there are a couple of things to consider before merge:

1. Whether we completely remove `babel` dependency, or keep it around. Up to you, I'm probably leaning towards remove. (See comment.)
2. The content width in the editor defaults to the "wide" view. I think it should default to narrow, as this matches the default template (with sidebar widgets), and works better when editing on smaller screens/windows. (See screenshot below.)

Screen Shot 2020-11-03 at 4 27 26 PM

Pre-approving (those fixes should be straightforward), feel free to ping me for another review if you like ✋

package.json
Aljullu: Good point! Removing them sounds like the best option to me, it makes the codebase simpler and decreases the maintenance burden. I have removed babel completely in https://github.com/woocommerce/storefront/pull/1525/commits/9dec992aeba91ec85127e2a0662810d245f678cf, we can always install it back if we need it in the future.

Aljullu:

> The content width in the editor defaults to the "wide" view.

Good catch! It definitely makes more sense defaulting to the narrow width. Should be fixed in https://github.com/woocommerce/storefront/pull/1525/commits/26f80a9b46d199843854b047e82c230534914a63.

I will wait merging this PR until tomorrow, so it targets Storefront 3.0 instead of 2.9, which is to be released tomorrow.

PR #1553, opened 2020-12-04
Fix implode parameters order.

### LINKED ISSUE #1551: php8 compatibility - implode parameter ordering
implode() for historical reasons could accept its parameters in either order (For consistency with explode()) however, it is deprecated not to use the documented order of arguments. 

I was running [PHP8 compatibility sniff](https://github.com/PHPCompatibility/PHPCompatibility) over the child themes and noticed the following in storefront core

> Passing the $glue and $pieces parameters in reverse order to implode has been deprecated since PHP 7.4 and is removed since PHP 8.0; $glue should be the first parameter and $pieces the second

Affecting the following line https://github.com/woocommerce/storefront/blob/trunk/inc/nux/class-storefront-nux-starter-content.php#L659

budzanowski:
Fixes #1551

Since PHP8 the alternative parameters order is not supported inside `implode`

### Testing

1. Install storefront
2. Install the starter content
3. Starter content taxonomies should still work.

@haszari I am targeting this directly to the release branch. When the release will be over this will get merged back to the trunk. Such a simple change so I did not want to go the cherry-picking road.

### DIFF
--- a/inc/nux/class-storefront-nux-starter-content.php
+++ b/inc/nux/class-storefront-nux-starter-content.php
@@ -656,7 +656,7 @@ if ( ! class_exists( 'Storefront_NUX_Starter_Content' ) ) :
 				$args['hide_empty'] = false;
 
 				// List of categories to display.
-				$args['ids'] = implode( $product_cats, ',' );
+				$args['ids'] = implode( ',', $product_cats );
 			}
 
 			return $args;


opr (approved):
After testing this out, I see the storefront theme in the customizer with no errors and the correct sample content. Looks good to me! 👍🏼 
Screenshot 2020-12-04 at 14 25 42 (2)

haszari:
> I am targeting this directly to the release branch. When the release will be over this will get merged back to the trunk. Such a simple change so I did not want to go the cherry-picking road.

Thanks for highlighting this in PR description @budzanowski – good to know! As long as the release branch gets merged back we should be good :)

PR #1471, opened 2020-09-19
👌 IMPROVE: Remove deprecated code (#1373)

nielslange:
Fixes #1373

Remove deprecated code.

### How to test the changes in this Pull Request:

1. When running a WooCommerce version < 2.3, the `add_to_cart_fragments` filter will no longer work.
2. When running a WooCommerce version < 2.5, the `star_rating` script will no longer work.
3. When running a WooCommerce version < 3.3, the `products_per_page` script will no longer work.
4. When running a WooCommerce version < 3.3, the `loop_shop_columns` filter and
the `woocommerce_before_shop_loop` and `woocommerce_after_shop_loop` actions will no longer work.

### Changelog

> Dev – Removed code for unsupported versions of WooCommerce (3.3 and earlier) #1373

PR #1567, opened 2021-01-04
Remove code targeting unsupported versions of WooCommerce

### LINKED ISSUE #1373: remove unused/deprecated code targeting older versions of WooCommerce
We've recently [clarified and updated the versions of WooCommerce that are supported for Storefront](https://docs.woocommerce.com/document/storefront-requirements-version-support/). There may be code targeting older versions which is no longer needed. Removing this code could simplify the code base and reduce the amount of code that needs to be shipped & parsed.

- Search code for `WC_VERSION` constant to find relevant code and safely remove as appropriate.
- There may be other ways to find out of date js/css code to remove also.

Follow up issue from discussion in https://github.com/woocommerce/storefront/pull/1368#discussion_r435688925

mikejolley:
This PR implements some of the changes from #1471 (not all of them since some of the removed methods are still being used, such as the infinite scroll ones).

I've included WC Version support headers in the readme.txt so this is easier to keep track of. I've set the min supported version to 4.2 which was the [min when this issue was logged](https://github.com/woocommerce/storefront/issues/1373). We could increase further but I didn't want to set it to a much later version in a fix release. 

Closes #1471
Fixes #1373 

Smoke test under supported versions of WC (4.2+).

### Changelog

> Fix – Removed legacy code targeting unsupported versions of WooCommerce. Storefront currently supports 4.2+.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -4,6 +4,8 @@ Requires at least: 5.3.4
 Tested up to: 5.6.0
 Stable tag: 2.9.0
 Version: 3.0.0
+WC requires at least: 4.2
+WC tested up to: 4.8
 License: GPLv2 or later
 License URI: https://www.gnu.org/licenses/gpl-2.0.html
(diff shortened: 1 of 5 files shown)


opr (approved):
Code changes look decent, and did a full run through from setup on a fresh site through to purchase using storefront, no errors and it all worked well.

PR #1569, opened 2021-01-04
Update engines in package.json

### LINKED ISSUE #1521: Update node/npm versions in package.json.engines
https://github.com/woocommerce/storefront/blob/f6db3d25597e246ae52e375e2fa362da634efa00/package.json#L62-L65

Very very old min versions. Should probably update to LTS.

I wonder if there's a bot that can help maintain these, I think I was assuming RenovateBot would notice this!

mikejolley:
I've just used the engines the blocks plugin uses; I tested a build and it seems ok. Will let the tests run to confirm.

Closes #1521


Aljullu (approved):
LGTM

PR #1576, opened 2021-01-07
Use proper escape function in comments template

donnapep:
The comment template was using an incorrect escape function for the `datetime` attribute in the comments template.

This PR updates the code to use `esc_attr` for the `datetime` attribute instead of `esc_html`.

1. View a blog post that has some comments.
2. Ensure the comment date is displayed and that the `datetime` attribute is set.

### Changelog

> Fix - Use proper escape function for `datetime` attribute in comments template. #1576


Aljullu (approved):
LGTM, thanks for the fix @donnapep!

PR #1581, opened 2021-01-12
Use proper escape functions in several templates

Aljullu:
Fixing some warnings about not using the proper escape functions that I received when releasing 3.2.0.

Changes are quite safe, but they can be tested as follows:

1. `comments.php`: go to Settings > Discussion and set the _Break comments into pages with_ option to 5. Then, add 6 comments to a post. **Verify:** the aria-label of `#comment-nav-above` has text.
2. `inc/storefront-template-functions.php`: in Appearance > Customize > Menus add a new menu to the secondary menu and publish it. **Verify:** the aria-label of `.secondary-navigation` has text.
3. `inc/nux/class-storefront-nux-admin.php` in a new site, install WooCommerce and activate it. After that, install Storefront. (Otherwise, in an existing install set `storefront_nux_dismissed` option to `0`). Go to Appearance > Storefront. **Verify.** the _Skip_ button is displayed correctly.
imatge

### Changelog

> Updated several escape functions.

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -37,7 +37,7 @@ if ( post_password_required() ) {
 		</h2>
 
 		<?php if ( get_comment_pages_count() > 1 && get_option( 'page_comments' ) ) : // Are there comments to navigate through. ?>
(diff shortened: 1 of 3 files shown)


mikejolley (approved):
👍🏻

PR #1574, opened 2021-01-06
Remove custom WooCommerce font usage

mikejolley:
Storefront was using the icon sets from WooCommerce core not by including them in Storefront, but by trying to reference them from the plugin directory. As pointed out in #1519, the plugin directory is not constant and can be changed, therefore breaking icons.

Whilst we could have looked for a method of making the path variable, considering how much usage there is of these icons in Storefront, it didn't seem worth the effort, especially considering how we also include FontAwesome for icons.

To that end, I removed the reference to the two icon fonts and replaced them with FontAwesome. There was a note inline saying that "FontAwesome didn't work for star icons", but I ran into no problems using FontAwesome in this context.

For the "note" icon, FontAwesome notice icon seems more appropriate. See the screenshots below.

Fixes #1519

### Screenshots

Before:
Screenshot 2021-01-06 at 16 09 40

After:
Screenshot 2021-01-06 at 16 10 51

1. Check a product page; confirm star ratings are displayed correctly, and the rating form is still functional
2. Go to a product with no reviews, an empty cart, or trigger an error notice and confirm the "notice" icon is displayed as pictured above.

### Changelog

> Fix – Replaced WooCommerce and Star icon usage with FontAwesome.

PR #1588, opened 2021-01-15
Fix missing icons

### LINKED ISSUE #1582: Checkout block: error notice icon is missing
## To Reproduce
1. Remove all available shipping methods from WooCommerce > Settings > Shipping.
2. Go to the Checkout block and scroll to the Shipping step.
3. Notice the notice icon is missing:
imatge

Maybe a regression from #1574?

### LINKED ISSUE #1586: Conflict with WooCommerce Blocks, Star Rating Icons Missing
## Describe the bug
WooCommerce blocks fail to load star rating icons when Storefront is active.

The blocks attempt to load `font-family: star;` which appears to be available as a .woff, .eot, and .ttf under`wp-content/plugins/woocommerce/assets/fonts/`

Other themes do not appear to be having issues.

Possible regression related to https://github.com/woocommerce/storefront/pull/1574
cc: @mikejolley @opr @Aljullu 

Discussed internally with @GeoJunkie and @fevered.

Isolating the problem:

## To Reproduce
1. Create a page using WooCommerce's product blocks.
2. Ensure at least one product that appears in the block has a rating.
3. View the front-end of the page.

## Screenshots
image

### WordPress Environment
Tested across WordPress.com, Pressable, Jurassic Ninja, and SiteGround.

Various infrastructures but all running WooCommerce 4.9 and Storefront 3.3.0

Aljullu:
Fixes #1582.
Fixes #1586.

This PR partially reverts the changes in #1574 (cc @mikejolley). Even though in theory we could update Storefront so it doesn't use any font family from WooCommerce, some extensions (at least, WC Blocks) rely on those fonts being loaded in the page. So I don't think we can remove them.

In order not to regress #1519 I moved the font loading to the PHP side.

#### Bug #1586

1. Verify stars in rated products are correctly displayed in PHP-based grid blocks (ie: Products by Category), All Products block and Review blocks.

imatge

#### Bug #1586

1. Force WC to show an error, you can do that in the Checkout block submitting the form with an empty address.
2. Verify the icon in the notice is displayed.

imatge

#### Bug #1519

1. Rename your `/wp-content/plugins/` folder to `/wp-content/modules/`.
2. Add these lines to `wp-config.php`:
```PHP
define( 'WP_PLUGIN_DIR', dirname(__FILE__) . '/wp-content/modules' );
define( 'WP_PLUGIN_URL',  'https://{yoursite.com}/wp-content/modules' );
```
3. Verify icons are displayed across the store.

### Changelog

> Fix missing icons in WC Blocks and some other parts of the UI.

### DIFF
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -11,31 +11,6 @@
 @import "../../../node_modules/susy/sass/susy";
 @import "../sass/vendors/modular-scale";
 
-// Star font, FontAwesome doesn't work :(
-@font-face {
-	font-family: star;
-	src: url(../../../../../plugins/woocommerce/assets/fonts/star.eot);
-	src:
-		url(../../../../../plugins/woocommerce/assets/fonts/star.eot?#iefix) format("embedded-opentype"),
-		url(../../../../../plugins/woocommerce/assets/fonts/star.woff) format("woff"),
-		url(../../../../../plugins/woocommerce/assets/fonts/star.ttf) format("truetype"),
-		url(../../../../../plugins/woocommerce/assets/fonts/star.svg#star) format("svg");
-	font-weight: 400;
-	font-style: normal;
-}
-
-@font-face {
-	font-family: WooCommerce;
-	src: url(../../../../../plugins/woocommerce/assets/fonts/WooCommerce.eot);
-	src:
-		url(../../../../../plugins/woocommerce/assets/fonts/WooCommerce.eot?#iefix) format("embedded-opentype"),
-		url(../../../../../plugins/woocommerce/assets/fonts/WooCommerce.woff) format("woff"),
-		url(../../../../../plugins/woocommerce/assets/fonts/WooCommerce.ttf) format("truetype"),
-		url(../../../../../plugins/woocommerce/assets/fonts/WooCommerce.svg#WooCommerce) format("svg");
-	font-weight: 400;
-	font-style: normal;
-}
-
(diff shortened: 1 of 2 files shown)


mikejolley:
@Aljullu Is it possible to fix both cases? I.e use fontawesome in storefront (because we use that font in other places and it makes sense to me) but load the old font for bw compat with extensions?

Aljullu:
> @Aljullu Is it possible to fix both cases? I.e use fontawesome in storefront (because we use that font in other places and it makes sense to me) but load the old font for bw compat with extensions?

Sounds good to me. Do you think that's blocking this PR? In my opinion it could be handled in a separate issue so this PR can still be merged to fix the relevant issues.

mikejolley:
@Aljullu  Unless I'm mistaken we'd just need to add the PHP code from this PR and not revert the fonts.

Aljullu:
@mikejolley good point. I was still able to reproduce #1582, but it should be fixed in a3d4369545e17f44b0b1979112fdd7df411a5dd5.

mikejolley:
All working except one thing which I'm not sure is related or not: 

Screenshot 2021-01-20 at 16 42 01

These are not quite aligned. All products block.

Aljullu:
Good catch @mikejolley! I think this is an issue with the All Products block, I opened a PR in WC Blocks with a more in-depth explanation: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3717.

mikejolley (approved):
👍🏻

PR #1600, opened 2021-02-01
Remove maximum-scale from viewport meta tag

### LINKED ISSUE #1599: Remove maximum-scale from viewport meta tag
Storefront defines `maximum-scale=2.0` in its viewport meta tag, but this is bad for accessibility. It limits how much users can zoom in the page.

Some browsers have an option to ignore it, but it would be better to completely remove it from the theme.

The relevant code line:

https://github.com/woocommerce/storefront/blob/2618a3119d93e7624bc3ebafcbb88be7a8e695c2/header.php#L14

Siddharth1010:
Fixes #1599 : Removed maximum-scale from the meta tag in `header.php`

### DIFF
--- a/header.php
+++ b/header.php
@@ -11,7 +11,7 @@
 <html <?php language_attributes(); ?>>
 <head>
 <meta charset="<?php bloginfo( 'charset' ); ?>">
-<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=2.0">
+<meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
That was fast @Siddharth1010, thanks for the fix! Testing well and code changes look good. :ok_hand: 

For reference, these are the testing steps I followed for Firefox Android:

1. Go to the browser settings > Accessibility > uncheck Zoom in all websites (copy might be slightly different because I'm not using it in English).
2. Reload the page and try to zoom in.
3. Verify you can zoom more than 2x.

PR #1615, opened 2021-02-23
Fix #1611 - Handheld-footer and notice only hidden on smaller devices

### LINKED ISSUE #1611: Handheld Footer Bar Hidden on Input Focus
When focusing on the quantity input the mobile footer bar disappears. 
Whats the reasoning for this behavior?

The CSS rule that is causing this:
https://github.com/woocommerce/storefront/blob/5a3619d4dbed265dbeab3ce64b2f9e211d7531b4/assets/css/woocommerce/woocommerce.scss#L239-L244

### To Reproduce
Focus on quantity input,

meirroth:
### Issues

1. Handheld footer bar is hidden when focusing on input fields. (#1611)
This behavior was intended for mobiles that open the on-screen keyboard when an input is focused, so the available vertical space is reduced considerably.
However on modern phones that have a big enough screen, or on devices with an external keyboard this behavior is not necessary at all.

2. Store notice was not being hidden, see screenshots below:
    <details> 
      <summary>Screenshots</summary>
       Initial and focused state.<br>
       
    </details>

### Fixes
1. As suggested by @Aljullu, moving the style rule inside a media query based on max-height (320px).
2. Adds the store notice message class to the same rule.

### Screenshots

Before and after on iPhone 5/SE landscape mode

### DIFF
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -236,10 +236,14 @@
 	}
 }
 
-.sf-input-focused {
+@media screen and (max-height: 320px) {
 
-	.storefront-handheld-footer-bar {
-		display: none;
+	.sf-input-focused {
+
+		.woocommerce-store-notice,
+		.storefront-handheld-footer-bar {
+			display: none;
+		}
 	}
 }
 


Aljullu (commented):
Thanks for opening this PR @meirroth, it looks good to me.

I will let somebody else from the team to give the final approval to this approach. @senadir will you take a look?

senadir (approved):

PR #1650, opened 2021-04-19
Fix potential error that can occur when product is null on sticky add to cart

wywarren:
Sometimes product pages can be presented without a global product as a result of plugins.

Returns out of the sticky add to cart function if the product object is null.

We have a certain configuration of plugins that causes this. It may be an edge case in a vanilla installation.

1. Go to a single product page
2. Scroll down until the add to cart button is out of view
3. Confirm the sticky cart shows as normal

### Changelog

> Ensure the product global exists in the sticky add to cart function


mikejolley:
Thanks for your contribution @wywarren. I think we can expand the testing instructions to just check that the functionality has no regressions in regular usage. I've gone ahead and added to your comment above.

mikejolley (approved):
Made a small change to cover cases where product is false too, but good to go 👍🏻

PR #1653, opened 2021-04-20
Remove skip-link-focus-fix.js

Ismail-elkorchi:
skip-link-focus-fix is only applicable in IE11, but it gets loaded even on the +95% of browsers which will never use it.

Now that [Matt is supportive of dropping IE11 support in WP Core](https://make.wordpress.org/core/2021/03/04/discussion-dropping-support-for-ie11/#comment-40870), I think it is the best time to remove it.

1. Open any StoreFront page in Safari (which is the only major browser using webkit).
2. Tab to skip link.
3. Press Enter to activate it.
4. Make sure keyboard focus has moved to the first [focusable area](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area).
5. Repeat this process for Opera and the other major browsers.

### Changelog

> Enhancement – Remove skip-link-focus-fix.js. https://github.com/woocommerce/storefront/pull/1653


nerrad:
We're definitely open to (and eventually will) remove IE11 only fixes, however, it's too premature for us to do so at this time. The [current IE11 phase out plan](https://make.wordpress.org/core/2021/04/22/ie-11-support-phase-out-plan/) starts with WordPress 5.8 release and _with a focus on admin side support initially_. Since our products still support earlier versions of WordPress we can't completely drop IE11 support until we bump up our minimum requirements.

I definitely appreciate the initiative here (and am looking forward to the day we can completely drop these IE11 workarounds!) but it'll be a while before we can merge pull requests like this.

Ismail-elkorchi:
WordPress 5.8 was already released, moreover, the [WooCommerce developper blog state that](https://developer.woocommerce.com/2021/05/24/developer-advisory-plan-to-remove-ie-11-support-in-woocommerce/) IE11 support will be fully removed by WooCommerce 5.6, which will be released on August 10th.

In light of recent events, can this pull request be reopened and merged?

nerrad:
I've re-opened this for reconsideration. However, I'd like to see some testing instructions in the pull request description please - especially since the previous conditionals in the code also included `webkit` and `opera` user agent checks, so on the surface, the existing code doesn't look like it's addressing only IE11.

nerrad:
It also appears the branch will need rebased on trunk.

Ismail-elkorchi:
Thank you for opening this PR for reconsideration.

>  I'd like to see some testing instructions in the pull request description please

Of course, I can do that.

> the previous conditionals in the code also included webkit and opera user agent checks, so on the surface, the existing code doesn't look like it's addressing only IE11.

The skip link focus fix used in storefront is an old implementation that targeted old browsers and was first used in 2013 by `_s` in https://github.com/Automattic/_s/pull/139. Then, it was updated in https://github.com/Automattic/_s/pull/1053 to reflect a fix introduced in Twenty Seventeen [by this ticket](https://core.trac.wordpress.org/ticket/38604) after discovering that this code doesn't work in IE11.  In this ticket, the code targeting older versions of webkit and Opera was removed, and only IE11 was checked. In short, all the browsers have fixed the bug that this code is targeting, and was only needed for IE.

mikejolley (approved):
Safari is working. I think it makes sense to remove this. If a site really wants support for older, unsupported browsers, I think it would be trivial to add a script back. Let's make sure to explain this in the release notes.

opr:
If you need to support older browsers, and want to include [`skip-link-focus-fix.js`](https://raw.githubusercontent.com/woocommerce/storefront/012999e187d78ef06fb96010096676779cc2f4ad/assets/js/skip-link-focus-fix.js) on your site you may download the file from the above link and add it to a directory accessible by your website. You may then follow the instructions in the [Theme Developer Handbook](https://developer.wordpress.org/themes/basics/including-css-javascript/) to enqueue the script on the front-end of your site.

PR #1654, opened 2021-04-21
Remove Jquery Pointer Events Polyfill (PEP)

Ismail-elkorchi:
By the time Jquery Pointer Events Polyfill was introduced (https://github.com/woocommerce/storefront/pull/982) the Pointer events API was not supported in all major browsers. This is not the case today, as the Pointer events API is now supported in the majority of current browsers, which makes this polyfill unnecessary.

> PEP development ceased in 2018. A few minor maintenance patches have been pushed since then, and the latest version (0.5.3) has been published to npm. However, native support for Pointer Events is relatively good in most browsers at this point - see [caniuse: Pointer Events](https://caniuse.com/pointer). This project is now archived.

Source : https://github.com/jquery-archive/PEP#project-status

> However, while PEP may now be deprecated, the future of Pointer Events themselves is looking good, with the native API now supported in the majority of current browsers (see [caniuse.com/pointer](https://caniuse.com/pointer)). For this reason, unless a project specifically targets older browser versions, we would strongly encourage developers to stop including PEP and to instead rely solely on native Pointer Events.

source : https://openjsf.org/blog/2021/02/10/pointer-events-polyfill-pep-enters-emeritus-status-at-the-openjs-foundation/

### Changelog

> Enhancement – Remove Jquery Pointer Events Polyfill (PEP). https://github.com/woocommerce/storefront/pull/1654


nerrad:
Thanks for the PR! Before we can consider this can you please add some testing notes around what this change impacts and verify you have tested your PR? Thanks!

mikejolley:
We can pull the testing instructions from #982 for this since when testing against current browsers this polyfill should no longer be required. Given the Polyfill is deprecated, I vote we remove it.

**Testing Instructions**

- Create a menu with nested items (dropdowns)
- Using a mouse, you should be able to hover the top item to reveal the menu and click on an item to navigate to a page.
- Using a touch screen device, when tapping on the top item, the first screen should open the dropdown. Tapping again on the top item should redirect to the page. Note you need to be looking at the site in desktop mode, not handheld. Use a larger device like an iPad or request desktop site in mobile safari.

mikejolley (approved):
Looks good. Tested Android and Safari IOS touch events and they function normally.

PR #1668, opened 2021-04-30
Accessibility: add aria-label on breadcrumbs

radex02:
Improves accessibility of the breadcrumbs

Added aria-label="breadcrumbs" on the breadcrumb's \<nav>


mikejolley:
Thanks for contributing @radex02 - this looks like a useful enhancement. I did push a fix to fix the 'textdomain' which needs to be wrapped in quotes, and an escaping function to ensure it's safe.

In future, it would be great to also include testing instructions and confirm your PR is tested. In this case there would have been a syntax error due to the textdomain.

**Testing instructions**

1. View page on the frontend with a breadcrumb, for example, the shop page
2. Inspect source and confirm the nav element has an aria label.

Thanks

PR #1672, opened 2021-05-04
Hide spinner when Payment Request button is disabled

### LINKED ISSUE #1539: Add new way to block UI during checkout for payment request buttons
Add the ability to block different parts of the UI differently during checkout - one way for the checkout form, and another way for the payment request buttons.

More details here:
https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1509#issuecomment-819803665

This fix should be built on top of this PR:
https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1509

Once the PR for this fix is approved, merge PR #1509 to `trunk`, and then merge the PR for this issue to `trunk`. This should be done in one go so that we can release this feature in the next Stripe release with the correct UI for blocking the checkout.

asumaran:
Fixes: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/1539

We are adding an additional "disabled" state to the Payment Request button in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1542

The behavior for the "disabled" state will be exactly the same as with the "blocked" state with the only difference being the "spinner" icon won't be displayed.

This PR hides the icon only when the "wc_request_button_is_disabled" is present. This way it won't affect previous versions of the theme even when using older versions of the `woocommerce-gateway-stripe` plugin.

### Screenshots

Blocked state

Screen Shot 2021-05-03 at 10 24 18 PM

Disabled state

Screen Shot 2021-05-03 at 10 24 48 PM

Change introduced here is pretty straightforward. Testing it locally will require to install `woocommerce-payment-setup` and set up Stripe in dev mode. Then checkout to this PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1542 then trigger the changes manually via JS like in the screenshots above. Let me know if you need detailed steps. I wasn't able to use `npm run wp-env` with WC Stripe.

### DIFF
--- a/assets/css/base/icons.scss
+++ b/assets/css/base/icons.scss
@@ -615,6 +615,16 @@ a.remove {
 	}
 }
 
+.wc_request_button_is_disabled {
+
+	.blockUI {
+
+		&::before {
+			content: none;
+		}
+	}
+}
+
 .woocommerce-pagination {
 
 	.next,
--- a/assets/css/base/icons.scss
+++ b/assets/css/base/icons.scss
@@ -618,6 +618,7 @@ a.remove {
 .wc_request_button_is_disabled {
 
 	.blockUI {
+		cursor: not-allowed !important;
 
 		&::before {
 			content: none;


mikejolley:
> Testing it locally will require to install woocommerce-payment-setup

@asumaran sorry, what is this? Do I just need to install the stripe extension?

The changes seem fine, I will test once the above is cleared up :)

asumaran:
@mikejolley I'm sorry. I meant `woocommerce-gateway-stripe`. That is to be able to see the Payment Request button like in the screenshots above.

mikejolley:
@asumaran It worked as described, but you do still get the loading animation on the cursor on hover. Is this intentional? Does BlockUI need to be used when the button is in a disabled state?

asumaran:
@mikejolley thanks for review.

> but you do still get the loading animation on the cursor on hover

Not intentional. The cursor is added via JS by the jQuery BlockUI plugin. I've submitted a [small change](https://github.com/woocommerce/storefront/pull/1672/commits/541199816d92ce4031074cc54402c32e33ab3475) to use a more suitable cursor when the button is disabled.

mikejolley (approved):
Thanks for taking care of the cursor. Marking approved.

PR #1690, opened 2021-05-20
check if Storefront_NUX_Admin_Inbox_Messages_Customize is defined bef…

senadir:
Gravity forms settings page seems to load certain hook twice, most precisely, `admin_notices` action, this caused `class-storefront-nux-admin-inbox-messages-customize.php` to be required twice and `Storefront_NUX_Admin_Inbox_Messages_Customize` to be defined twice.

Other possible solutions include:
- importing `class-storefront-nux-admin-inbox-messages-customize.php` outside the hook.
- Understanding why Gravity forms is triggering this action twice and fixing it.

I opted for checking if the class is already defined.

### Testing instructions
- Download a testing copy of gravity forms premium from this private slack thread p1621431332030900-slack-C7U3Y3VMY
- Go to Gravity Forms settings page
- It should load fine


senadir:
Hiding widespace would make this a 2 line PR.

frontdevde (approved):
Thanks for looking into this! I tested the change and can confirm it fixes the underlying issue (see screenshots below).

The coding standards test was failing, so I pushed a minor change to fix that.

LGTM 👍

|Before|After|
|-|-|
|Screenshot 2021-05-21 at 11 17 22|Screenshot 2021-05-21 at 11 17 48|

PR #1691, opened 2021-05-21
Image optimisation with ImageAlpha (png8+alpha) and ImageOptim.

lowwebtech:
Just did some simple image optimization to reduce image and plugin weights.

PNG images are converted to PNG8 (256 colors often less).

- visually compare compressed PNG and JPG images with previous versions

### Changelog

> image optimization (imagealpha & imageoptim)


opr (approved):
The changes look fine to me, there are no noticeable differences in the image quality, and the smaller size is definitely a good thing! Thanks for this 👏🏼

Issue #1585, opened 2021-01-14
Full-width Elements Have Margins on Mobile View


robertocyrino:
## Describe the bug
If you add a full width element to a site using the latest **Storefront** theme, 3.3.0, it will show margins on both sides for mobile view.

- [x] I have deactivated other plugins and themes and confirmed this bug occurs when only WooCommerce + Storefront theme are active.
- [x] I can reproduce this bug consistently using the steps below.

## To Reproduce
1. Add an Image block to a new page, set it to full width;
2. Make sure the page's template is also set to Full Width, and save/publish the changes;
3. Open the created page on a browser with at least 1064px width. Verify that the image is full width.
4. Switch to browser responsive mode and reduce the screen size to 1063px width or less. Verify that there are margins on both sides.

## Screenshots

![https://d.pr/i/826xAu](https://d.pr/i/826xAu+) 
Full Size: https://d.pr/i/826xAu

## Expected behavior
Full width elements should stay full width in mobile view, as it happens with different themes such as Twenty Twenty-One.

### Browser Environment

Firefox 84.0.2 (64-bit), macOS 10.15.7

### WordPress Environment

WordPress 5.6
WooCommerce 4.9.0
Storefront 3.3.0

3648664-zen / 25542112-hc

mikeicode:
For the example above it appears the CSS which makes the blocks full width is within a media query causing this issue (not sure if that's intentional):

![https://d.pr/i/NX6Dj8](https://d.pr/i/NX6Dj8+)

We added this to the site to resolve:

```
.alignfull {
  margin-left: calc(50% - 50vw);
  margin-right: calc(50% - 50vw);
  width: auto;
}

```

Aljullu:
Thanks for filling this issue @robertocyrino and for the work-around @mikeicode.

> it appears the CSS which makes the blocks full width is within a media query causing this issue (not sure if that's intentional)

I'm confused by this as well, I think it should be safe to remove it and let styles be applied in all viewport sizes. Do you want to open a PR fixing this? :slightly_smiling_face:

Aljullu:
I'm assigning high priority because this is an issue in a core Gutenberg functionality and I imagine it has a big impact, but it looks like this has been working like that for several versions, so it's not a recent regression.

9ete:
@robertocyrino @mikeicode @Aljullu  PR to resolve issue open here: https://github.com/woocommerce/storefront/pull/1713

albarin:
[...]

PR #1713, opened 2021-06-21
remove-full-width-image-mobile-margin-1585: remove full width image block margins on mobile

9ete:
Moved styles targeting entry content full width images from applying only on 1024+ to apply on all viewport sizes.  This fixes the issue of image blocks marked as full width not going full width on mobile.

**Steps to test the fix** (originally posted here: https://github.com/woocommerce/storefront/issues/1585#issue-786076052):
- Add an Image block to a new page, set it to full width;
- Make sure the page's template is also set to Full Width, and save/publish the changes;
- Open the created page on a browser with at least 1064px width. Verify that the image is full width.
- Switch to browser responsive mode and reduce the screen size to 1063px width or less. Verify that there are no margins on both sides.

**Before Fix**
Screen Shot 2021-06-23 at 11 23 54 AM
Screen Shot 2021-06-23 at 11 23 35 AM

**After Fix**
Screen Shot 2021-06-23 at 11 21 45 AM
Screen Shot 2021-06-23 at 11 21 32 AM

### DIFF
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -19,24 +19,25 @@
  * Front-end only styles
  */
 .hentry .entry-content {
+
+	.storefront-align-wide.page-template-template-fullwidth-php & .alignfull,
+	.storefront-align-wide.storefront-full-width-content & .alignfull {
+		margin-left: calc(50% - 50vw);
+		margin-right: calc(50% - 50vw);
+	}
+
 	// Global
 	@media ( min-width: $container-width ) {
 
 		.storefront-align-wide.page-template-template-fullwidth-php &,
 		.storefront-align-wide.storefront-full-width-content & {
 
-			.alignfull,
 			.alignwide {
 				width: auto;
 				max-width: 1000%;
 				padding-left: ms(2);
 				padding-right: ms(2);
 				clear: both;
-			}
-
-			.alignfull {
-				margin-left: calc(50% - 50vw);
-				margin-right: calc(50% - 50vw);
 			}
 
 			.alignwide {


9ete:
@opr for issue: https://github.com/woocommerce/storefront/issues/1585

opr:
Hi @hawkeye126 thanks for making this PR! It would be really useful if you could edit the PR description to include a summary of what you changed and to add some testing steps. If you ping me back when you've done this I'll be able to take a look.

9ete:
@opr Done, thanks for the feedback and reviewing!  Please let me know if I can provide any further detail.

opr (approved):
Thanks @hawkeye126 - I've tested this and it looks to be working well! 👏🏼 Congratulations on your first PR to Storefront! 🥳 

I also ran `npm run lint:css` on my local machine and it passes. The job fails in GitHub due to permission errors I believe. More information on why the linting job failed: https://github.com/actions/first-interaction/issues/10

vanderwijk:
There seems to be an issue with this on my site. After updating to 3.8.0 my full width cover block is no longer spanning the full width of the page. This is because the width is set to 100% instead of auto.

screenshot

Adding auto width fixes this for me:

`	.storefront-align-wide.page-template-template-fullwidth-php & .alignfull,
	.storefront-align-wide.storefront-full-width-content & .alignfull {
		margin-left: calc(50% - 50vw);
		margin-right: calc(50% - 50vw);
		width: auto;
	}`

Destrega:
Just updated to 3.8.1 but the problem is still the same for me.

Changing the width from 100% to auto for .wp-block-cover fixes it for me.

PR #1755, opened 2021-08-09
Prevent search widget styles leaking into the Search block when added to a widget area

### LINKED ISSUE #1745: Search block has an overlaping icon when added to the sidebar
## Describe the bug
There is a styling issue between the Search block included in WordPress and Storefront styles when the block is added to a widget area.

## To Reproduce
1. With WordPress 5.8, go to Appearance > Widgets.
2. Add the Search block into the sidebar.
3. Go to the frontend, and notice there is a magnifier icon which overlaps the label:
imatge

It looks like the issue is coming from some styles from the widget which are leaking into the block?

Aljullu:
Fixes #1745.

Some selectors targeting the Search widget were also affecting the Search block when added to a widget area.

1. With WordPress 5.8, go to Appearance > Widgets.
2. Add the Search block into the sidebar.
3. Go to the frontend, and verify there isn't a magnifier icon which overlaps the label:

| Before | After |
| --- | --- |
| Screenshot before | Screenshot after |

4. Try toggling some attributes of the Search block and verify it's displayed as expected in the frontend.

### Changelog

> Fix – Several visual issues when a Search block was added to a widget area.

### DIFF
--- a/assets/css/base/_base.scss
+++ b/assets/css/base/_base.scss
@@ -1535,7 +1535,7 @@ button.menu-toggle {
 .widget_search,
 .widget_product_search {
 
-	form {
+	form:not(.wp-block-search) {
 		position: relative;
 
 		input[type="text"],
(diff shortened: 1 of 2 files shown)


opr (approved):
Works well, thanks Albert!

PR #1762, opened 2021-08-18
Add width:auto to full width images

### LINKED ISSUE #1761: Full width block broken in 3.8.0
When using a full width cover block, this is no longer spanning the full width of the browser.

I have been able to reproduce this on two different sites. See screenshots below.

This issue is caused by a missing 'width: auto' from PR #1713 

V3 7 0
V3 8 0

See also this support topic by a different user https://wordpress.org/support/topic/white-space-on-right-side-blocks-are-not-full-width/

opr:
This is required because on the Cover block, the image was set to 100% width which meant it didn't stretch the entire width of its parent if the image was smaller than the parent's width.

Changing this will cause the image to stretch, respecting the full width setting from the Cover block.

See this comment: https://github.com/woocommerce/storefront/pull/1713#issuecomment-900976617
Fixes #1761 

### Screenshots

| Before | After |
|---|---|
| image | image |

1. Add a cover block to a page using the `Full width` template on your site.
2. Set the cover to be full width too.
3. Visit the page and ensure the block covers the full width of the page.
4. Also use a regular image block and ensure it respects the various width settings.

### Changelog

> Fix - The image shown in the Cover block will now respect the full-width setting.

### DIFF
--- a/assets/css/base/gutenberg-blocks.scss
+++ b/assets/css/base/gutenberg-blocks.scss
@@ -24,6 +24,7 @@
 	.storefront-align-wide.storefront-full-width-content & .alignfull {
 		margin-left: calc(50% - 50vw);
 		margin-right: calc(50% - 50vw);
+		width: auto;
 	}
 
 	// Global


frontdevde (approved):
Tested as per testing instructions and can confirm this fixes the issue compared to `trunk`.

|Before|After|
|-|-|
|Screenshot 2021-08-18 at 12 35 08|Screenshot 2021-08-18 at 12 36 20|

PR #1674, opened 2021-05-05
Update build tools

mikejolley:
This PR introduces changes to the build process, including using node for the build process instead of grunt, our standard set of linting tools, and a GitHub workflow that replaces the Travis Job.

- Adds our suite of tooling for code styling/linting and closes #1656
- Updates the license information which closes #1669
- Adds browserlist config instead of defining manually which closes #1497
- Removes grunt and uses npm instead for building which closes #1433
- Moving away from grunt has fixed #1662 

Note, several CSS and JS files were modified due to the linting checks (things like spacing and indentation).

**Testing Instructions**

1. Confirm tests have passed; we have linting checks and the e2e task running via GitHub Workflow
2. `composer install && npm install && npm run build`
3. Confirm the build is successful and a zip archive is built. You can compare this to the last release zip - I used a diff tool and there are only minor differences due to updated tools. There should be no dev or build files within the release archive. **There should be** css files in assets/css and min.js files in assets/js.
4. Install the release archive and smoke test visual to ensure there are not style regressions.
5. Smoke test the JS functionality, for example, go to a product page and scroll so the add to cart button is out of view. This will cause sticky cart to appear. Or in admin, if you go to customizer for the first time you see a working guided tour.

When this PR is approved, it should also get the ready to merge label, just like the Blocks repo.

PR #1763, opened 2021-08-18
Remove `.gitattributes` file and add excluded files to `composer.json`

### LINKED ISSUE #1760: Since 3.6.0 - Release Source code archives missing build/compilation files
## Describe the bug
Since Release 3.6.0, the .zip/.tar.gz Source code assets, are missing all of the build/compilation related files (i.e. Grunt, Composer, Sass, etc...).

For example, download the 3.6.0 [Source Code (zip)](https://github.com/woocommerce/storefront/archive/refs/tags/3.6.0.zip), and it only contains the .php, .txt, and image (.jpg, .png., .svg, etc) files.

Download the 3.5.1 [Source Code (zip)](https://github.com/woocommerce/storefront/archive/refs/tags/3.5.1.zip) and it contains the hidden .github directory, plus all of the other grunt, composer, sass, etc. files.

## To Reproduce
1. Go to [https://github.com/woocommerce/storefront/releases/tag/3.6.0](https://github.com/woocommerce/storefront/releases/tag/3.6.0)
2. Download and extract [https://github.com/woocommerce/storefront/archive/refs/tags/3.6.0.zip](https://github.com/woocommerce/storefront/archive/refs/tags/3.6.0.zip) or [https://github.com/woocommerce/storefront/archive/refs/tags/3.6.0.tar.gz](https://github.com/woocommerce/storefront/archive/refs/tags/3.6.0.tar.gz)
3. Open the extracted directory.
4. Voilà!

## Screenshots
NA

## Expected behavior
I expect all the build/compilation stuff to be in the Source Code archives.

FYI - Specifically, I use Grunt to build all the sass stuff for each release, so that I can merge the latest css changes with my custom child theme's css code and I then minify and serve only my own .css assets on my prod websites (to reduce HTTP requests) I'm a stickler for performance :-D, and this also helps to keep my custom/child theme css structurally/syntactically inline with Storefront css targeting the correct elements/tags, etc...

This was easier before Storefront switched to minified/Sass style sheets somewhere around Storefront 2.x.x ( I can't remember ), but it's easy enough to build the sass stuff with Grunt since then to get the desired beautified css files with comments.

If you could put all this back into the archives (particularly the latest 3.8.0 one) that'd be much appreciated, I'm about 3 months behind on my theme updates... 8-P

Thanks!

### Browser Environment
NA

- Platform: NA
- Browser(s): NA

### WordPress Environment
NA

opr:
Semi-Fixes #1760 - there will be steps to do after this PR is merged.

To summarise the issue, in #1674 we added a `.gitattributes` file. This excluded files from the _source archives_ while the intention was to exclude them only from the _built archives_. This PR removes that file, as it's not necessary. We're using composer to generate the built archive in the `postbuild` script. Files that should not be included in this archive have been added to the `composer.json` file.

We will probably need to go back and tag point releases to allow users to download source archives for the affected versions.

1. Run `npm run build` and inspect the generated archive. Ensure it doesn't contain development or config files such as `.gitignore`.
2. Upload the archive to a JN site and ensure it works properly.
3. General smoke test of Storefront functionality

### Changelog

> Fix – Resolved an issue where source archives downloaded from GitHub did not contain all files.


Aljullu (approved):
Followed the testing steps and did some smoke testing and everything is looking good.

PR #1769, opened 2021-08-26
Update Woo's default colour

jobthomas:
The Woo default colour has changed from #96588a to #7f54b3. This PR replaces all instances of the old colour with the new colour. 

To test:

* Install Storefront from scratch -- look for the purple
* Specifically in the customizer -> color the default/accent colour will be updated


nielslange (approved):
LGTM!

PR #1771, opened 2021-08-26
Add support for menu description.

### LINKED ISSUE #1595: Support for menu description.

Currently, Storefront theme does not support menu descriptions.

Add the support for menu descriptions.

## Additional context

1. Head to wp-admin >> Appearance >> Menus
2. In the "Screen options" at the top, tick the "descriptions" checkbox.
3. Add descriptions to some menu items. That won't work.

Screenshot
Screenshot direct link: https://d.pr/i/oROi8H

Reported in 26823683-hc

nielslange:
**Note**

[WordPress 3.0](https://wordpress.org/support/wordpress-version/version-3-0/) introduced Custom Menus which also allows [adding a menu description](https://wordpress.org/support/article/appearance-menus-screen/#rearranging-configuring-menu-items). This PR aims to utilize this feature, so that Storefront users can show menu descriptions.

Fixes #1595

### Screenshots

<table>
<tr>
<td>Before:
<br><br>

#1595-before
</td>
<td>After:
<br><br>

#1595-after
</td>
</tr>
</table>

1. Go to `/wp-admin/nav-menus.php`.
2. Click on screen options in the upper-right corner and activate the option `Description` within `Show advanced menu properties`.
3. Create a custom menu.
4. Add 2-3 menu items with a description.

### Changelog

> Add support for menu description.


senadir (commented):
While the code is looking good, I'm not so sure about the implementation. Other themes seems to distinguish between the item and extra text, 2021 for example.
image.
For storefront, it seems to ruin base line balance, and it's hard to distinguish between the description and the actual link (from link color, to text size and placement).
image

Can we run this by someone else first?

nielslange:
@senadir: Thanks for your initial review.

@vivialice: Would you have the capacity to spin up a design for showing the menu descriptions when using the Storefront theme? In https://github.com/woocommerce/storefront/pull/1771#issue-980412780, you can see how the menu descriptions are looking currently.

tomasztunik:
Made so the items align like @senadir mentioned it is on 2021 (1.). Also updated styles to follow typographical scale used in the theme for secondary items like with cart counter (2.)
Screenshot 2022-03-21 at 19 38 50

Would be great to wrap this up but got a question - do you think adding `vertical-align: top` to allow this alignment as default could potentially affect child-themes or usage in the wild? In theory if people had a lot of variable length menu item names that break into multiline this could slightly alter default behaviour where it would align to the top like visible now rather than to the bottom like it was on @senadir's screenshot.

nielslange:
@tomasztunik Thanks for following up on this issue. Looking at your screenshot, I wonder if the color contrast of the menu description is sufficient, given that [WordPress aims to comply to the WCAG 2.0](https://wordpress.org/about/accessibility/).

tomasztunik:
Bumped the contrast on description and cart count to have 4.5x contrast in the default state.

Screenshot 2022-03-21 at 21 31 12

gigitux:
Hi!
I'm not sure that we can release this PR. I noticed that we have the same problem with the alignment when I display the menu as `Secondary Menu`. 

image

My suggestion is to revert the PR and fix this problem too. What do you think?

nielslange:
> My suggestion is to revert the PR and fix this problem too. What do you think?

@gigitux First of all, nice catch. I totally missed the submenu descriptions. I wonder if reverting is less effort, though, than spinning up an additional PR to cover the submenu as well. You know what, I'll quickly spin up another PR. You can then release Storefront either with both PRs, if mine gets reviewed fast enough. Or you revert this PR and the next porter releases these features next week.

PR #1785, opened 2021-09-18
fix/add-to-cart-button-text-breaking: fix: prevent product list products' button text from wrapping mid word, mentioned in https://github.com/woocommerce/woocommerce/issues/30200

9ete:


9ete:
This commit addresses the WooCommerce issue #30200 https://github.com/woocommerce/woocommerce/issues/30200

Aljullu:
Thanks for opening this PR @hawkeye126! I will wait reviewing it until there is a resolution of the discussion in https://github.com/woocommerce/storefront/issues/1787.

For the next steps, I would suggest if you could add testing instructions to the PR (even if you copy & paste it from the issue you mentioned). Adding before & after screenshots is also very useful. Also, you don't need to include a link to the issue you are fixing in the PR title, you can do so in the PR description, that helps keeping PR titles shorter. :slightly_smiling_face:

9ete:
Thanks for the feedback @Aljullu, that all makes sense to me.  I will add the additional info via the other PR I open for this task as you advised.

Aljullu:
Closing as this is superseded now by #1797.

PR #1812, opened 2021-10-07
Add .nvmrc file with version 12.21.0 specified

### LINKED ISSUE #1811: Add .nvmrc file to prevent broken builds.
With the latest version of Node the `npm run build` command has some issues creating the required minified versions of required JS files and produces a broken build. This occurred during a deployment, you can read more on this incident report pca54o-2gc-p2

To mitigate this issue from happening I'd like to see a `.nvmrc` file specifying a compatible version of Node (`12.21.0`)

## Describe alternatives you've considered
N/A.

## Additional context
pca54o-2gc-p2

tjcafferkey:
Fixes https://github.com/woocommerce/storefront/issues/1811

### Description
With the latest version of Node the npm run build command has some issues creating the required minified versions of required JS files and produces a broken build. This occurred during a deployment, you can read more on this incident report pca54o-2gc-p2

### Solution
To mitigate this issue from happening I've added a `.nvmrc` file specifying a compatible version of Node (12.21.0)

1. Install [nvm](https://github.com/nvm-sh/nvm). Skip this step if you already have it installed.
2. Using terminal navigate to your local `storefront` repository root
3. Check that upon navigating into the repo that nvm installs Node 12.21.0
4. Double check the correct version by running `node -v`

### Changelog

> Add – .nvmrc to ensure developers are running a compatible version of node. #1811


ralucaStan (commented):
Hey @tjcafferkey,
I've tested this PR and the version did not change automatically for me because my terminal was not set up to run `nvm use` when a `.nvmrc` file is detected. So this means that this will not work for everybody.

I've also notice that we have [another node version, 14.18.1](https://github.com/woocommerce/storefront/blob/trunk/package.json#L102) specified under engines in package.json. This means that we recommend installing the project with 14.18.1, but we also switch with your changes to a lower version. Does the build not work with 14.18.1?

I also think you should specify in the description of the PR & ticket what version do you mean by "With the latest version of Node" for a better future reference.

tjcafferkey:
Hey @ralucaStan 

>my terminal was not set up to run nvm use when a .nvmrc file is detected. So this means that this will not work for everybody.

Yep, this won't work for everyone. It's not intended to be a fix for every scenario but more of a preventative measure. For example, this would have fixed the issue for me seeing as I do use `nvm`. You can also find `.nvmrc` files in a lot of A8c repos that are in place to fix similar issues (Woo Blocks, Calypso & Jetpack to name a few).

>I've also notice that we have another node version, 14.18.1 specified under engines in package.json. This means that we recommend installing the project with 14.18.1, but we also switch with your changes to a lower version.

I specified the lower version because I have verified this works as its the version I fixed a broken release with which was caused by an incompatible Node version. If you'd rather leave this issue/PR open until the next release where we can smoke test again with version `14.18.1` that's fine also.

ralucaStan:
Hei @tjcafferkey,
I've actually noticed that the node version in package.json is [updated automatically by renovate](https://github.com/woocommerce/storefront/pull/1817). Considering we have issues with the latest version maybe we should prevent this update. What do you think?

I see that the last node version we actually specified was 14.16.1, in @mikejolley's [PR](https://github.com/woocommerce/storefront/pull/1674/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R101), so I expect the release to actually work with this version?!

>  If you'd rather leave this issue/PR open until the next release where we can smoke test again with version 14.18.1 that's fine also.

I think we should be consistent with the node versions we enforce/recomment. We can bring this up for discussion with the team as well.

ralucaStan:
I've [prevented](https://github.com/woocommerce/storefront/pull/1826#issuecomment-951740570) renovate from updating Node until we make a decision here.

tomasztunik:
It is possible to prevent npm from running outside of allowed engines but it's disabled by default. 

We'd need to ensure `package.json` engines and `.npmrc` have the same version set and add `engine-strict=true` flag in `.npmrc` file to prevent npm from running on non supported version. By default it only shows a warning that's easy to skip over.

Question I have is should we go with v12 or v14 as @ralucaStan mentions @tjcafferkey?

tomasztunik (approved):
Updated the version to use the one which is used by CI and specified in engines

PR #1897, opened 2022-02-07
Fix editor alignment issues

### LINKED ISSUE #1320: Gutenberg trunk: several blocks are left aligned on wide screens
## To Reproduce
1. Get Gutenberg master.
2. Add a paragraph or a list block + some other blocks.
3. In a wide viewport, notice paraghraph and list are left-aligned but the other blocks are centered.

This happens because `<p>` and `<ul>` have left and right margins set to 0. We would want to set the left and right margins to `auto` to fix this. For the list, this is a bit more tricky because it needs a left margin of `3em`, we could try to make it a padding instead.

However, I think it makes sense to wait before fixing this until an official version of Gutenberg is released. Twenty Twenty seems to be affected as well.

## Screenshots
imatge

DustinHartzler:
Fixes #1320 

There were a couple of heading blocks and blockquote blocks that did not align with the rest of the text in the editor. Added `rem` to a couple of the calculations and change the blockquotes to have a left/right margin of `auto`

### Screenshots

**Here's the before:**
alignment-before

**After**
alignment-after

1. Build the theme and install
2. Add different heading types (H1, H2, etc) and blockquotes
3. The should all have the same left justification

- No flows are affected

### Changelog

> Fix – Left justified heading and blockquote elements. #1320


assets/css/base/gutenberg-blocks.scss
```diff
@@ -299,7 +299,7 @@
 
 	// Blockquote
 	.wp-block-quote {
-		margin: 0 0 ms(2);
+		margin: 0 auto;
```
tomasztunik: You have removed accidentally bottom margin here causing blockquote to "stick" to the next block that follows it. Also let's change it to `margin-bottom: ms(2);` to avoid it overriding .wp-block sytles for horizontal margins

assets/css/base/gutenberg-editor.scss
```diff
@@ -306,7 +306,7 @@ table {
 }
 
 .wp-block[data-align="wide"] {
-	max-width: ms(19);
+	max-width: ms(19) rem;
```
tomasztunik: This would produce invalid css and _fix it_ by doing so - it would fall back to default values that produce "better" ux but not matching storefront frontend. We'll need to remove it.

Screenshot 2022-02-09 at 14 56 07

assets/css/sass/utils/_variables.scss
```diff
@@ -22,7 +22,7 @@ $info:              #3d9cd2;
 // layout sizes
 $desktop:           768px;
 $handheld:          568px;
-$container-width:   ms(18);
+$container-width:   ms(18) rem;
```
tomasztunik: Same case here, this will produce invalid CSS and calculations.

tomasztunik (changes requested):
Hey Dustin, thanks for picking this one up! It was small but definitely annoying bugger! I've added couple notes to enable us to merge this. Would appreciate if you could have another look at it!

I believe the suggestion from https://github.com/woocommerce/storefront/issues/1320#issuecomment-868549528 in `/assets/css/base/gutenberg-editor.scss:L303-310` might be the best bet here. I believe we should be able to get around it without using !important there. As it was suggested as a mean to override current broken settings.

Was looking at changing modular scale base from em to rem but altering it will affect vertical rhythm of the pages too much as it would affect every ms() function usage. On top of it there seems to be a bug and even though ms() function should accept custom bases locally so we use it only for max-width on blocks it refuses to work 🤷 

I have confirmed that this issue is local to storefront theme.

DustinHartzler:
Thanks @tomasztunik for the feedback. CSS isn't my area of expertise, but this bug was driving me crazy!

I've tested the above changes on one of my test sites and the headers are now left justified with the rest of the text 🎉

assets/css/base/gutenberg-editor.scss
```diff
@@ -303,12 +303,11 @@ table {
 
 // Main content width
```
gigitux, line 304: What do you think if we add a comment about this fix here? For example in this way:
```suggestion
// Main content width
// These two rules are copied from https://github.com/woocommerce/storefront/issues/1320#issuecomment-868549528
```

See these numbers so specific, it is a rare thing, so I guess that it is better to give a little bit of context

gigitux (changes requested):
Thanks for working on this! 🙏 

I suggest to add a comment about the fix and at this moment there are lint errors. This is the command for fix them:

```
npm run lint:css-fix
```

sunyatasattva (approved):
Personally, I feel the solution is really convoluted with those weird numbers. Thanks @gigitux for adding the comment at least it explains some context.

I still wish there was a better way, but as a patch, I suppose it is fine.

PR #2040, opened 2022-09-08
Add actions to single product pagination and sticky add-to-cart template functions

LuigiPulcini:
At the moment, the `storefront_single_product_pagination` and `storefront_sticky_single_add_to_cart()` template functions do not feature any action hooks. The proposed changes add some extra calls to `do_action()` inside those two functions so that it would be possible not only to add any extra HTML markup before or after those elements (which could be even achieved by hooking to the appropriate actions those functions are hooked to) but also detect whether something is occurring inside those functions. For example, if a plugin needed to apply some filters but limit their adjustments only inside either the single product pagination or the single product sticky add-to-cart, the following logic could be used:

```php
add_filter( 'woocommerce_single_product_image_html', 'my_change_to_product_thumbnail_html' );
function my_change_to_product_thumbnail_html( $html ) {
    if ( did_action( 'storefront_before_single_product_pagination' ) > did_action( 'storefront_after_single_product_pagination' ) ) {
        // we are inside the pagination element
        // so do something here to alter the markup of the product image
        $html = my_function_to_alter_the_product_thumbnail_html_markup( $html );
    }

    return $html;
}
```
### Changelog

> New - Added actions to single product pagination
> New - Added actions to sticky single add-to-cart

### DIFF
--- a/inc/woocommerce/storefront-woocommerce-template-functions.php
+++ b/inc/woocommerce/storefront-woocommerce-template-functions.php
@@ -770,23 +770,45 @@ if ( ! function_exists( 'storefront_single_product_pagination' ) ) {
 			return;
 		}
 
+		do_action( 'storefront_before_single_product_pagination' );
+
 		?>
 		<nav class="storefront-product-pagination" aria-label="<?php esc_attr_e( 'More products', 'storefront' ); ?>">
-			<?php if ( $previous_product ) : ?>
+			<?php
+			if ( $previous_product ) {
+
+				do_action( 'storefront_before_single_product_pagination_previous' );
+				?>
+
 				<a href="<?php echo esc_url( $previous_product->get_permalink() ); ?>" rel="prev">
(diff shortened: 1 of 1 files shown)


danieldudzic:
Hi @LuigiPulcini,

We really appreciate you taking the time and creating this PR.

There is still an investment of time needed by our team to consider the impact of merging your PR to Storefront. Because at this time, your pull request does not fall into one of [the criteria "buckets"](https://github.com/woocommerce/storefront/blob/trunk/STOREFRONT_STATUS.md), we will go ahead and close it.

LuigiPulcini:
Thanks for the explanation and for adding a link to the criteria "buckets" – as you call them.

To the best of my knowledge, **adding** `do_action` calls to template functions that do not currently make use of any template because the HTML they output is hardcoded into them should not have any consequences and produce no compatibility issues. So, I am not quite sure what kind of _impact_ or _investment of time_ you might be referring to.

In any case, thank you for considering my PR request.

PR #2054, opened 2022-10-12
Prevent infinite loop on prev/next product navigation

### LINKED ISSUE #2053: Prevent infinite loop on prev/next product navigation
`Storefront_WooCommerce_Adjacent_Products::get_product()` has no execution time limit. So it may take huge execution time or enter infinite loop.
In my environment, it causes by [Simple Custom Post Order](https://wordpress.org/plugins/simple-custom-post-order/).

Memo a previous adjacent ID for prevent infinite loop.

chrono-meter:
Fixes #2053.
Storefront_WooCommerce_Adjacent_Products::get_product() has no execution time limit. So it may take huge execution time or enter infinite loop.

In my environment, it causes by [Simple Custom Post Order](https://wordpress.org/plugins/simple-custom-post-order/) ( [\SCPO_Engine::scporder_previous_post_where()](https://plugins.trac.wordpress.org/browser/simple-custom-post-order/tags/2.5.6/simple-custom-post-order.php#L472) and [\SCPO_Engine::scporder_next_post_where()](https://plugins.trac.wordpress.org/browser/simple-custom-post-order/tags/2.5.6/simple-custom-post-order.php#L500)).
This is not StoreFront's bug, and it's only occurred in some environments. But it's more better experience by adding infinite loop blocker.

### DIFF
--- a/inc/woocommerce/class-storefront-woocommerce-adjacent-products.php
+++ b/inc/woocommerce/class-storefront-woocommerce-adjacent-products.php
@@ -81,18 +81,20 @@ if ( ! class_exists( 'Storefront_WooCommerce_Adjacent_Products' ) ) :
 
 			$product               = false;
 			$this->current_product = $post->ID;
+			$infinite_loop_blocker = array();
 
 			// Try to get a valid product via `get_adjacent_post()`.
 			// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
-			while ( $adjacent = $this->get_adjacent() ) {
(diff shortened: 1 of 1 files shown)


danieldudzic:
Hi @chrono-meter,

We really appreciate you taking the time and creating this PR.

I'd recommend reporting the issue to the plugin developer so that they can investigate whether there's a way to adjust their code not to cause conflict.

The conditional check added could cause a loop break out too early meaning there may be valid adjacent products that won't be surfaced.

Also, there is still an investment of time needed by our team to consider the impact of merging your PR to Storefront. Because at this time, your pull request does not fall into one of [the criteria "buckets"](https://github.com/woocommerce/storefront/blob/trunk/STOREFRONT_STATUS.md), we will go ahead and close it.

Issue #2077, opened 2022-12-04
Serve Source Sans locally - avoid law suits


koppor:
## Is your feature request related to a problem? Please describe.

In Germany, it is forbidden to fetch Fonts from Google. If one doesn't follow this rule, one has to play (at least) 300€ per case. See https://ll-ip.com/aktuelles/raag-kanzlei-dikigoros-nikolaos-kairis/ for details.

## Describe the solution you'd like

Storefront should ship the font in its distribution.

## Describe alternatives you've considered

There is a hint at https://wordpress.org/support/topic/css-font-display-how-to-add/ to create a child theme. It seems, however that 300+ lines of code need to be copied - only for replacing one font.

dinhtungdu:
Hi @koppor,

Thanks for your report. I acknowledge the issue of Google Fonts and GDPR and I can confirm Storefront is loading Source Sans directly from Google Fonts as you described.

> Storefront should ship the font in its distribution.

Thanks for your suggestion. Our team will need to discuss this before taking any action. Because Storefront is now in maintenance mode, the process can take a while.

In the meantime, you can use a plugin to make your site GDPR-compliant: https://wordpress.org/plugins/local-google-fonts/. Can you please test that plugin to see if it solves your issue?

koppor:
Thank you for the quick response. The plugin works great, thank you for the hint!

Aljullu:
Heads-up that in https://github.com/woocommerce/storefront/issues/845#issuecomment-386379581 there is a suggested snippet that might also work. There is also [this filter](https://github.com/woocommerce/storefront/blob/32bb2c33ae10291658b37eb8e068d016352e0031/inc/class-storefront.php#L385) that might be useful.

PR #2080, opened 2022-12-13
add font swap to google font css api

dowands:
Add font swap to google font css display:swap to resolve "Ensure text remains visible" problem on web.dev

### DIFF
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -391,6 +391,7 @@ if ( ! class_exists( 'Storefront' ) ) :
 			$query_args = array(
 				'family' => implode( '|', $google_fonts ),
 				'subset' => rawurlencode( 'latin,latin-ext' ),
+				'display' => 'swap',
 			);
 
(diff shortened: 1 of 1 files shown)


imanish003:
Hey @dowands, Thank you for taking the time to create this PR; we appreciate your contribution 🙏🏻

I had an internal discussion with the team regarding this. A store might prefer to display the content as soon as possible and replace the font when downloaded (`font-display: swap`). Still, other stores might prefer to block rendering until the font has completely downloaded `font-display: auto/block`. Therefore we don't want to hard-code `swap` as `font-display` value. 

Also, FYI, In future, we plan to remove Google Fonts because Google Fonts cause a GPDR issue for EU websites. You can find more info about it in issue #2077. 

One alternative way to change `font-display` value could be to unregister Google Fonts with the methods proposed in #2077 & then register the fonts directly with a CSS file, so you can add the font-display property as you want. 

I will close this PR for now, but we can continue our discussion in the comments. Thanks again for your contribution 🙂

PR #2086, opened 2023-01-12
Hide the single product page quantity input, when the product is sold individually.

### LINKED ISSUE #2085: Single Product Page > Maybe Hide Quantity Selector
In WooCommerce 7.2 we changed the way [product quantity selectors render](https://github.com/woocommerce/woocommerce/pull/34282). As part of this change, we began rendering the quantity selector even when a product is set to be sold individually ... however, there was [considerable resistance to this change](https://github.com/woocommerce/woocommerce/issues/36007).

Subsequently, we decided to restore the previous behavior. The preferred way of doing this, though, is now via CSS rather than setting the input type to hidden.

We have already made the necessary change within [WooCommerce core](https://github.com/woocommerce/woocommerce/pull/36350). However, since Storefront deqeues the default WooCommere stylesheet we need to echo the same change within Storefront.

## Describe alternatives you've considered
We don't necessarily need to do anything, but the volume of [recent feedback](https://github.com/woocommerce/woocommerce/issues/36007) we received over in the core WooCommerce repo suggests making a change to Storefront may be wise.

## Additional context
This change is only applicable to the single product page and not to other pages.

barryhughes:
If a product is set to be sold individually, let's hide the quantity input.

Fixes #2085.

### Screenshots

| Regular Product | Sold Individually |
| --- | --- |
| test-product-regular | test-product-soldindiv |

The above shows how things will look once this change is applied (the quantity input will be hidden, if the product is set to be sold individually).

1. Configure a product to be sold individually.
2. View the corresponding single product page (and please test this against WooCommerce 7.2.0 or higher).
3. With this change, the quantity input should be hidden.
4. Without this change, the quantity input will still be visible.
5. Confirm the quantity input **always** renders for products that are **not** set to be sold individually.

💡 When testing this branch, be sure to rebuild the stylesheets using `npm run build:css`. You may also need to clear your browser cache to see the change.

### Changelog

> Fix – Hide the single product page quantity selector if the product is sold individually. #2085

### DIFF
--- a/assets/css/woocommerce/woocommerce.scss
+++ b/assets/css/woocommerce/woocommerce.scss
@@ -715,6 +715,10 @@ ul.products,
 			font-size: ms(-1);
 			margin-top: 1em;
 		}
+
+		&.sold-individually .quantity {
+			display: none;
+		}
 	}
 }
 


barryhughes:
Closing: we've reverted the corresponding change in WooCommerce (when the input should be hidden, it will once again be set to `type="hidden"`) and so no longer require a CSS solution.

PR #2100, opened 2023-06-09
[DRAFT] Replace mini-cart widget with mini-cart block and add support for template parts.

nerrad:
**Note: this PR should not be merged as is**

This is a very limited proof of concept to aid discussions about having Storefront default to using the Mini-Cart block instead of the Woo Core mini-cart widget. 

The team has been working on some improvements around the use of `get_refreshed_fragments` API in Woo Core, however the mini cart widget will still enqueue this script (and the resulting ajax calls). Implementing some variation of what is suggested in this PR will accomplish a number of things:

- Enable the default state of the theme with a more performant implementation of the mini-cart behaviour via the block version.
- Provide an example for other classic theme shops/developers for how to utilize the new Mini-Cart block in their themes.
- Via enabling template part support (in concert with the changes needed in [WooCommerce Blocks here](https://github.com/woocommerce/woocommerce-blocks/pull/9780)), provides a way for merchants/users of this theme to have more no-code options for customizing the behaviour and design of the Mini-Cart contents panel.

### Screenshots

The Mini-Cart block in action on the frontend:
CleanShot 2023-06-09 at 14 40 54

The template part:
CleanShot 2023-06-09 at 15 02 53

1.
2.
3.

## Additional Notes:

- There are some CSS/styling issues that would need to be handled. Storefront has some custom styles for blocks that impact some of the styles on the frontend. Note that the positioning of the block in the default area is out of alignment vertically.
- The editor WYSIWYG view for the template part does not match the actual rendering on the frontend. I suspect we'd have to either add `theme.json` support to the theme _or_ enqueue relevant styles in the editor context to achieve the parity.
- We'd have to decide how we'd roll this out for all users. For example, would it be something new users of Storefront default to, something that existing users opt-in to? Should we still leave the implementation of the widget available for folks to utilize?
- Merchants would not be able to edit the actual implementation of the block here. This isn't a significant change for users of the Storefront theme because they weren't able to do that beforehand anyways. 
- We'd need to test this with a Storefront child theme and consider what behaviour should exist here.

### Changelog

> Enhancement - Implement Mini-Cart block as a replacement for the Mini-Cart widget. Also enables block template part support for the theme.

### DIFF
--- a/header.php
+++ b/header.php
@@ -15,6 +15,13 @@
 <link rel="profile" href="http://gmpg.org/xfn/11">
 <link rel="pingback" href="<?php bloginfo( 'pingback_url' ); ?>">
 
+<?php
+// Need to call do_blocks() here to ensure any script/styles are enqueued in wp_head.
+global $storefront_block_mini_cart;
+// todo: could potentially implement better default attributes for the block.
+$storefront_block_mini_cart = function_exists( 'do_blocks' ) ? do_blocks( '<!-- wp:woocommerce/mini-cart /-->' ) : null;
+?>
+
 <?php wp_head(); ?>
 </head>
 
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -247,6 +247,13 @@ if ( ! class_exists( 'Storefront' ) ) :
 					),
 				)
 			);
+
+			/**
+			 * Add support for block emplate parts.
+			 *
+			 * @return void
+			 */
+			add_theme_support('block-template-parts');
 		}
 
 		/**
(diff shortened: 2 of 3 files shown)


nerrad:
Despite authoring this PR and not having any technical concerns about the implementation (other than followups around cleaning up the CSS), I’m not sure we should actually ship Storefront with these changes.

While it would be a good example for classic themes, there are a number of considerations we’d have to make around whether this is opt-in/opt-out, do we default new stores using Storefront to use the Mini-Cart block, what happens with child themes etc. The actual code I did is trivial, the product questions and work around answering those, supporting it after release etc, makes the effort questionable. 

However, I’m surfacing this so others can give their take if they disagree.

PR #2106, opened 2023-07-12
Add exclusion for `/vendor` directory in zip build

danielwrobert:
The `composer archive` command previously included the `/vendor` directory where the `.gitignore` file excludes it.

This adjusts the exclude array in the `composer.json` file to omit the `/vendor` directory from the zip generation.

See this internal conversation for additional context: p1689174493505199-slack-C7U3Y3VMY.

Props to @danieldudzic for discovering the discrepancy!

1. Run the `composer install` command which will generate a `/vendor` folder in the root directory.
2. Run `npm run build` and confirm the `storefront.zip` file is generated.
3. Confirm that the zip file is ~2.9 mb in size.
4. Open the generated zip file and confirm that there is no `/vendor` folder present in the root directory of the theme.

### Changelog

> Fix – add exclusion for /vendor directory in release zip build.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,12 +11,12 @@
 		"exclude": [
 			"!/assets",
 			"!/languages",
-			"!/vendor",
 			"!/style.css",
 			"!/style-rtl.css",
 			"/assets/css/sass",
 			"/assets/css/**/*.scss",
 			"/e2e",
+			"/vendor",
 			"composer.json",
 			"composer.lock",
 			"package.json",


Aljullu (approved):
LGTM, thanks for creating the PR, @danielwrobert!

I followed the testing steps and did some smoke testing with the generated ZIP and everything seems to be working well.

nefeline (approved):
👍 All good! Confirmed the `vendor` directory is not part of the generated zip file anymore.

PR #2113, opened 2023-08-01
Enqueue wc-cart-fragments on woocommerce pages

gigitux:
Fixes #2101

From WooCommerce Core 7.8, the `wc-cart-fragments` isn't enqueued by default anymore. This causes #2101. I followed the ["Best practices for the use of the “cart fragments” API" post](https://developer.woocommerce.com/2023/06/16/best-practices-for-the-use-of-the-cart-fragments-api/) to fix the issue.

Also, given that the PR is small, I bumped the tested version to 8.0.

### How to test the changes in this Pull Request:

1. Add a product to the cart
2. Go to the cart or the checkout page (shortcode or blocks versions)
3. Hover over the **Storefront header cart**. Ensure no popup is displayed showing the cart info
4. Update the cart (e.g., Add quantity, remove an item, add a coupon, etc.). Save changes if you are using the shortcode version of the cart and checkout pages. Ensure the **Storefront header cart** reflecting the current changes.

### Changelog

> Fix: Enqueue wc-cart-fragments script on woocommerce pages.

PR #2115, opened 2023-08-10
Load wc-cart-fragments-script when the widget cart is active

### LINKED ISSUE #2114: #2113 breaks refreshing the cart on non-WC cached pages
#2113 has an unintended consequence: it prevents the cart being refreshed on non-WooCommerce pages that are cached with a disk cache solution such as WP Super Cache. Until the 4.5.0 release, a cached page could potentially be served with the "wrong" cart information (the information from the user that triggered the page being cached) but that was not a problem since wc-cart-fragments would update the cart with the correct information almost immediately. This change breaks this for non-WooCommerce pages. Pages from WooCommerce do still get the correct cart information, but other pages (for example a contact page) do not anymore. This makes impossible to cache any non-WooCommerce page.

gigitux:
Fixes #2114

With #2113, we start to enqueue the `wc-cart-fragments-script` only on WooCommerce Pages, but the `Mini Cart Widget` can be loaded in all the pages of WordPress. So, I update the logic to ensure that the script is loaded when the `Mini Cart Widget` is active.

1. Install WP-Optimize - Clean, Compress, Cache plugin. 
2. Enable the cache.
3. Add the Mini Cart widget via the widget page (Appareance > Widget). Add in the header.
4. Create a post  and add the `On Sale Products` block (it is important that it is an old block that it is server side). Save it.
5. Visit the post.
6. Add the products in the cart.
7. Ensure that the Mini Cart widget is updated.
8. Refresh the page.
9. Ensure that the Mini Cart widget is loaded and shows the right product in the cart.

-
-

### Changelog

> Load  `wc-cart-fragments-script` when Mini Cart widget is active.

### DIFF
--- a/inc/class-storefront.php
+++ b/inc/class-storefront.php
@@ -378,7 +378,7 @@ if ( ! class_exists( 'Storefront' ) ) :
 		}
 
 		/**
-		 * Limit Cart Sync functionality to specific WooCommerce pages
+		 * Limit Cart Sync functionality to specific WooCommerce pages or when the Widget Cart is active.
 		 * More details: https://developer.woocommerce.com/2023/06/16/best-practices-for-the-use-of-the-cart-fragments-api/
 		 *
 		 * @param string $script_data The script data.
@@ -387,7 +387,7 @@ if ( ! class_exists( 'Storefront' ) ) :
 		 */
 		public function limit_cart_sync_to_wc_pages( $script_data, $handle ) {
 			if ( 'wc-cart-fragments' === $handle ) {
-				if ( is_woocommerce() || is_cart() || is_checkout() ) {
+				if ( is_woocommerce() || is_cart() || is_checkout() || is_active_widget( false, false, 'woocommerce_widget_cart', true ) ) {
 					return $script_data;
 				}
 				return null;


albarin (approved):
👍

A-Printer:
This PR fixes things with the cart widget, but not with the cart that is present in the menu, [here on the right of the screen](https://i0.wp.com/themes.svn.wordpress.org/storefront/4.5.0/screenshot.png?w=1144&strip=all).

gigitux:
> This PR fixes things with the cart widget, but not with the cart that is present in the menu, [here on the right of the screen](https://i0.wp.com/themes.svn.wordpress.org/storefront/4.5.0/screenshot.png?w=1144&strip=all).

Thanks for the feedback! Could you share some reproduction steps, please?

This is an attempt that I did and it seems that works correctly:


<video src=https://github.com/woocommerce/storefront/assets/4463174/08cb2c03-7c3c-44e7-8b5b-ba685fa5c9b3 />

webwit:
Try it on the homepage or other none woo page (and without mini cart) ?

A-Printer:
Here you can see that the cart is updating on the WooCommerce pages, but not on the Contact page:

https://github.com/woocommerce/storefront/assets/87852864/c330dbe1-05be-4896-89d3-82dc37140dfd

webwit:
Yeah this doesn't fix the cart at the top right as I also reported in #2114. I don't use the mini cart.
The top right cart is part of the storefront templates, that header is used all over my site. Ehm, just a layman, but that function to only include it on some pages based on conditions - if you're running storefront with that cart always in the header, what is the purpose?

A-Printer:
@gigitux I agree with @webwit, since by default the cart (not the widget) is part of the navigation and is present on every page, it should get refreshed with wc-cart-fragments-script on every page, no exceptions.

albarin:
Hi @webwit and @A-Printer, thanks for your suggestions! 🙌  
Could you have a look at this new PR https://github.com/woocommerce/storefront/pull/2116 and see if it fixes the issue for you? Thanks a lot!

webwit:
Hi Alba,

I'm out and didn't test, but that should do it as yesterday I outcommented the call to add_filter to limit_cart_sync_to_wc_pages and that fixed it for me. #2116 removes that and effectively does the same.

A-Printer:
@albarin I can confirm that #2116 does fix the issue. Thanks!

albarin:
Thank you both for your help! We'll create a new release as soon as possible.

PR #2138, opened 2024-01-16
Change shops to shop's as it needs an apostrophe

damiencarbery:
"shops appearance" - possessive so should be "shop's appearance". Or drop the s: "shop appearance" and avoid any questions about apostrophes.


inc/admin/class-storefront-admin.php
```diff
@@ -138,7 +138,7 @@ public function storefront_welcome_screen() {
 						<img src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/admin/welcome-screen/child-themes.jpg" alt="Storefront Powerpack" />
 
 						<p>
-							<?php esc_html_e( 'Quickly and easily transform your shops appearance with Storefront child themes.', 'storefront' ); ?>
+							<?php esc_html_e( 'Quickly and easily transform your shop\'s appearance with Storefront child themes.', 'storefront' ); ?>
```
Aljullu, line 141:
```suggestion
							<?php esc_html_e( "Quickly and easily transform your shop's appearance with Storefront child themes.", 'storefront' ); ?>
```

Aljullu (commented):
Thanks for opening this PR, @damiencarbery! Good catch with the typo. I suggest using double quotes `"` to wrap the string, this way we don't need to escape the single quote `'`. I left the code suggestion inline below. :point_down:

PR #2149, opened 2024-05-16
Ensure that the build has min.js files before generating the .zip file

gigitux:
This PR adds a sanity check step before generating the .zip file to proceed with the release.

When the right version of node isn't used, the *.min.js files aren't generated and this may create some issues on website. This PR ensures that the `min.js` files exist before creating the zip.

## Testing instructions:

1. Checkout this branch
2. Switch to the node 14 version.
3. Run `npm run build`.
4. Ensure that everything works correctly.
5. Switch to the node 20 version (or another version different from 14).
6. Run `npm run build`.
7. Expect to see an error.

### DIFF
--- a/validate-build.sh
+++ b/validate-build.sh
@@ -32,6 +32,7 @@ done < <(find "$base_dir" -type f -name "*.js" ! -name "*.min.js" -print0)
 
 # Exit with an error code if any .js file is missing its corresponding .min.js file
 if [ "$error_flag" -ne 0 ]; then
+  echo "Error: the ZIP could not be built because minified scripts are missing. Please ensure you are using the correct versions of NPM and Node.js."
   exit 1
 else
   echo "All .js files within $base_dir have corresponding .min.js files."
--- a/validate-build.sh
+++ b/validate-build.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Initialize the error flag
+error_flag=0
+
+# Define the base directory to check
+base_dir="./assets"
+
+# Check if the base directory exists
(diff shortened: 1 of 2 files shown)


validate-build.sh
```diff
@@ -0,0 +1,39 @@
...
+# Find all .js files within the base directory (excluding .min.js files)
+while IFS= read -r -d '' js_file; do
+  if ! check_for_min_js "$js_file"; then
+    error_flag=1
+  fi
+done < <(find "$base_dir" -type f -name "*.js" ! -name "*.min.js" -print0)
+
+# Exit with an error code if any .js file is missing its corresponding .min.js file
+if [ "$error_flag" -ne 0 ]; then
+  exit 1
```
Aljullu, line 36: What do you think about adding a troubleshooting message here, something like:

"Error: the ZIP could not be built because minified scripts are missing. Please ensure you are using the correct versions of NPM and Node.js."

Aljullu (approved):
Works like a charm! I could reproduce the issue on my end when running the build script with the wrong node version. I left one minor comment, but pre-approving.

validate-build.sh
gigitux: Added with [9aba910](https://github.com/woocommerce/storefront/pull/2149/commits/9aba9101a404fc769bfeab58abb94079d8e5fef1)

gigitux:
> Works like a charm! I could reproduce the issue on my end when running the build script with the wrong node version. I left one minor comment, but pre-approving.

Thanks for the review! I added the troubleshooting message! Great suggestion!

PR #2158, opened 2024-08-29
Fix: Rating block is showing "S" instead of star on Editor side.

imanish003:
Rating block is showing "S" instead of star on Editor side, for example on new posts or pages. This PR fixes the issue by enqueuing the WooCommerce font in the editor.

Fixes #

1. Add "All Reviews" block to a new post or page.
2. Verify that the rating block is displaying correctly.

| Before | After |
| --- | --- |
| image | image |

3. Publish the post or page, and verify that the rating block is displaying correctly on the frontend.

### Changelog

> Fix: Rating block is showing "S" instead of star on Editor side.


Aljullu:
@imanish003 given that this is affecting all classic themes, I wonder if we should fix it in WooCommerce core directly instead of making the changes in the themes themselves. I opened an issue in WC: https://github.com/woocommerce/woocommerce/issues/51139. What do you think?

imanish003:
@Aljullu Makes sense to me. We can close this PR 💯

imanish003:
I am closing this PR as we will fix it upstream in WooCommerce core.

PR #2202, opened 2025-12-09
fix: fix performance issue with adjacent product in 6.9

dinhtungdu:
Fixes https://github.com/woocommerce/woocommerce/issues/62233

- WordPress 6.9 adds an `AND p.ID` fallback in `get_adjacent_post()`. When two products share the same publish date and the adjacent one is hidden, the query could repeat/hang because the ID wasn’t swapped in that clause. This PR fix this issue by altering the clause with not only date but also the current product ID.

### How to test the changes in this Pull Request:
1. Use Storefront; open a product on the front end (e.g., `Polo`) and note its next product (e.g., `T-Shirt with Logo`).
2. Edit that next product (`T-Shirt with Logo`): set its publish date/time to exactly match the first product and set visibility to Hidden.
3. Reload the first product page; confirm it loads without timing out, and the prepared next/prev links do not point to the hidden product.

PR #2203, opened 2025-12-09
Storefront_WooCommerce_Adjacent_Products: Add progress guard to loop

mreishus:
Adds a small guard to break out of possible infinite loops in the `Storefront_WooCommerce_Adjacent_Products::get_product()` function.  This doesn't solve any current issues (the recent one is fixed in #2202), this is more of a fail-safe to make sure that infinite loops can't happen again, since SQL filtering can be brittle.

https://en.wikipedia.org/wiki/Defense_in_depth_(computing) 

### References
* https://core.trac.wordpress.org/ticket/64390
* #2202

Can undo the recent mitigations in #2202 and recreate an infinite loop by having WordPress 6.9 with three products in the same category, with the same post_date, and the middle one set to be hidden from catalog. This should also mitigate the same problem.

Additionally, regular back, forward arrows between products should continue to work.

### Changelog

> Enhancement - Protect the adjacent product query against future infinite loops originating from outside changes. #2203

### DIFF
--- a/inc/woocommerce/class-storefront-woocommerce-adjacent-products.php
+++ b/inc/woocommerce/class-storefront-woocommerce-adjacent-products.php
@@ -82,9 +82,16 @@ if ( ! class_exists( 'Storefront_WooCommerce_Adjacent_Products' ) ) :
 			$product               = false;
 			$this->current_product = $post->ID;
 
+			$seen = array();
+
 			// Try to get a valid product via `get_adjacent_post()`.
 			// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
 			while ( $adjacent = $this->get_adjacent() ) {
+				if ( isset( $seen[ $adjacent->ID ] ) ) {
+					break; // give up if we're not making progress
+				}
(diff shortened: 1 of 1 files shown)


inc/woocommerce/class-storefront-woocommerce-adjacent-products.php
```diff
@@ -82,9 +82,16 @@ public function get_product() {
 			$product               = false;
 			$this->current_product = $post->ID;
 
+			$seen = array();
+
 			// Try to get a valid product via `get_adjacent_post()`.
 			// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
 			while ( $adjacent = $this->get_adjacent() ) {
+				if ( isset( $seen[ $adjacent->ID ] ) ) {
```
dinhtungdu, line 90: Great care using `isset` here, thank you!

dinhtungdu (approved):
Thanks so much for the PR. I agree we should prevent the loop from happening again, since we're not the only ones touching the clause. :shipit:

PR #2205, opened 2025-12-11
Test: verify infinite loop fix for adjacent products

dinhtungdu:
- Add PHPUnit setup and regression tests for adjacent products fix introduced in https://github.com/woocommerce/storefront/pull/2202.
- Add GitHub Actions workflow to run PHP unit tests against stable/nightly WP.

### Screenshots
n/a

Check GH Actions or run the test locally.
1. npm install
2. composer install
3. npm run wp-env start
4. npm run test:php

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "storefront",
-			"version": "4.6.0",
+			"version": "4.6.2",
 			"license": "GPL-3.0+",
 			"devDependencies": {
 				"@woocommerce/eslint-plugin": "2.3.0",
--- a/tests/php/StorefrontWooCommerceAdjacentProductsTest.php
+++ b/tests/php/StorefrontWooCommerceAdjacentProductsTest.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * Tests for Storefront_WooCommerce_Adjacent_Products using WP_UnitTestCase.
+ *
+ * @package storefront
+ */
+
+class StorefrontWooCommerceAdjacentProductsTest extends WP_UnitTestCase {
+
+	/**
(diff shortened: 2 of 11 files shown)


package-lock.json
```diff
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "storefront",
-			"version": "4.6.0",
+			"version": "4.6.2",
```
kmanijak, line 9: Huh, that's weird, it's 4.6.2 on `trunk` and this branch seems up to date with `trunk`... 😅 not sure why GitHub shows it as diff.

tests/php/StorefrontWooCommerceAdjacentProductsTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+class StorefrontWooCommerceAdjacentProductsTest extends WP_UnitTestCase {
+
+	/**
+	 * Next product should skip hidden products with identical publish dates.
+	 */
+	public function test_next_product_skips_hidden_with_same_date() {
+		$date         = '2024-01-01 10:00:00';
+		$current_post = $this->create_product( $date );
+		$hidden_post  = $this->create_product( $date, true );
```
kmanijak: Nitpick: let's call it `hidden_next` similar to the next test where it's called `hidden_prev`

```suggestion
		$hidden_next  = $this->create_product( $date, true );
```

kmanijak (approved):
Super cool to see this test setup added and actual tests for the `Storefront_WooCommerce_Adjacent_Products` class!

I was able to run tests locally (and they pass!). I only left some nitpick comment and updated the testing steps with `composer install` (it wasn't there and couldn't run tests at first). But pre-approving.

Great job Tung! 🚀

PR #2211, opened 2026-05-07
Remove wp-scripts e2e toolchain

dinhtungdu:
Started as a dependency update to address the Axios security issue. However, I noticed that we only have that package because it's a dependency of `wp-script`, and we only use `wp-script` for E2E testing. Our E2E test suite has only one simple test that can be covered by our PHP test. So, I decided to remove the ET test and all of the wp-script dependencies.

Testing:
- `npm run lint:js` ✅
- `npm run test:php` ✅
- Verify the built is identical:
```sh
# Verify built assets are identical to trunk
TRUNK_DIR="$(mktemp -d)"
trap 'rm -rf "$TRUNK_DIR"' EXIT
git clone --branch trunk --depth 1 --single-branch . "$TRUNK_DIR" 2>/dev/null
(cd "$TRUNK_DIR" && npm ci --silent 2>/dev/null && npm run build:dev --silent 2>/dev/null)
npm ci --silent 2>/dev/null && npm run build:dev --silent 2>/dev/null
diff -rq assets "$TRUNK_DIR/assets" && \
diff -q style.css "$TRUNK_DIR/style.css" && \
diff -q style-rtl.css "$TRUNK_DIR/style-rtl.css" && \
echo "✅ No differences" || echo "❌ Differences found"
```

Changelog:
> Dev – Remove wp-scripts e2e tooling and cover the footer credit smoke test in PHP.


package.json
```diff
@@ -92,27 +72,29 @@
...
 		"rtlcss": "4.3.0",
+		"sass": "1.77.8",
+		"stylelint": "14.16.1",
 		"susy": "2.2.14",
 		"typescript": "5.5.4",
 		"uglify-js": "3.19.3"
 	},
+	"overrides": {
+		"axios": "^1.15.1"
+	},
```
gigitux: I think that we can remove the overrides if we remove the `wp-scripts` package - what do you think?


```suggestion

```

.github/workflows/e2e-tests.yml
gigitux, line 1: Deleting the E2E workflow also removes the only CI job that runs `npm run build`. The remaining workflows cover linting and PHP tests, but they won’t catch broken JS/CSS compilation or packaging regressions from this dependency change.

Could we keep a lightweight build validation job or move `npm ci && npm run build` into another existing workflow?

gigitux (approved):
I left few comments, but I'm pre-approving the PR. Thanks for working on this!

package.json
dinhtungdu: ah, this is left from the first step, sure, that's the point of this pr.

.github/workflows/e2e-tests.yml
dinhtungdu: yeah, makes sense

PR #2221, opened 2026-09-15
Remove dead .eot font references for star and WooCommerce icon fonts

### LINKED ISSUE #2176: star.eot and woocommerce.eot fonts don't exist anymore
## Describe the bug
Storefront theme contains links to two fonts [star.eot](https://github.com/woocommerce/storefront/blob/trunk/inc/woocommerce/class-storefront-woocommerce.php#L109) and [woocommerce.eot](https://github.com/woocommerce/storefront/blob/trunk/inc/woocommerce/class-storefront-woocommerce.php#L120) from WooCommerce that were removed in https://github.com/woocommerce/woocommerce/pull/31670. They produce 404 errors as reported [here](https://wordpress.org/support/topic/error-404-font-face/#post-18113361).

- [x] I have deactivated other plugins and themes and confirmed this bug occurs when only WooCommerce + Storefront theme are active.
- [x] I can reproduce this bug consistently using the steps below.

## To Reproduce
1. Notice 404 notices when browser tries to load missing .eot files

## Expected behavior
I would expect no 404 errors when loading the font

### Browser Environment
I'm not sure which browsers load .eot over .woff currently.

### WordPress Environment
WC 9.4
WP 6.6

Arslan-TR:
## What

WooCommerce removed `star.eot` and `WooCommerce.eot` in [woocommerce/woocommerce#31670](https://github.com/woocommerce/woocommerce/pull/31670). Storefront still registers `@font-face` rules pointing at those files, so every page load produces 404s for them ([reported here](https://wordpress.org/support/topic/error-404-font-face/#post-18113361)).

## Fix

Drop the `.eot` `src` entries from both `@font-face` blocks in `add_core_fonts()`. The `woff`/`ttf`/`svg` fallbacks already cover current browsers, so nothing regresses.

Fixes #2176


gigitux (approved):
I'm not able to reproduce the 404 error mostly because modern browsers generally skip EOT and load WOFF, but this is a good refactor. LGTM

